PR #798, opened 2026-07-11
fix(cli): hooks add resolves bundled dist/hooks layout

### LINKED ISSUE #797: hooks add: hookSourceDir() misses the bundled dist/hooks layout — always fails on npm installs
## Summary

`harness hooks add <name>` always fails on a bundled (npm/mise) install with:

```text
x Failed to add hook: Hook scripts not found: …/lib/node_modules/@harness-engineering/hooks
```

because `hookSourceDir()` in `packages/cli/src/commands/hooks/add.ts` only tries the **dev layout** path:

```ts
function hookSourceDir(): string {
  const d = path.resolve(__dirname, '..', '..', 'hooks');
  if (fs.existsSync(d)) return d;
  throw new Error(`Hook scripts not found: ${d}`);
}
```

In the published package the command is bundled into `dist/`, so `__dirname` is `…/@harness-engineering/cli/dist` and `../../hooks` resolves to `…/@harness-engineering/hooks`, which doesn't exist. The scripts actually ship at `dist/hooks/` (copied by `copy-assets.mjs`).

`hooks init` had this same bug and already carries the fix — `resolveHookSourceDir()` in `packages/cli/src/commands/hooks/init.ts` tries both layouts:

```ts
const candidates = [
  path.resolve(__dirname, '..', '..', 'hooks'), // dev layout
  path.resolve(__dirname, 'hooks'),             // bundled layout (copy-assets.mjs)
];
```

`add.ts` was never given the same treatment.

## Repro

```bash
mise use -g npm:@harness-engineering/cli@5.0.0   # or any npm install
cd some-harness-project
harness hooks add sentinel-pre
# x Failed to add hook: Hook scripts not found: …/@harness-engineering/hooks
```

## Impact

A project whose committed `.claude/settings.json` registers a hook (e.g. `sentinel-pre`/`sentinel-post`) that isn't installed locally has no working repair path: `hooks add` is exactly the documented remedy and it always throws on real installs. Meanwhile every Claude Code tool call spams non-blocking `Cannot find module …/.harness/hooks/sentinel-pre.js` hook errors. (Workaround: manually copy the scripts out of `…/@harness-engineering/cli/dist/hooks/`.)

## Fix

Share `resolveHookSourceDir()` between `init` and `add` (or duplicate the two-candidate probe in `add.ts`). PR incoming.

- CLI version: 5.0.0 (also present on `main` as of 2026-07-11)

bstevenski-capillary:
Closes #797.

## Problem

`harness hooks add <name>` always fails on published (npm/mise) installs:

```text
x Failed to add hook: Hook scripts not found: …/@harness-engineering/hooks
```

`hookSourceDir()` in `packages/cli/src/commands/hooks/add.ts` only probed the dev layout (`__dirname/../../hooks`). In the published package the command is bundled flat into `dist/`, so that resolves to the nonexistent `@harness-engineering/hooks` sibling; the scripts actually ship at `dist/hooks/` (via `copy-assets.mjs`).

## Fix

`init.ts` already carries the correct two-candidate probe (`resolveHookSourceDir()` — dev layout, then bundled layout). This PR exports it and uses it from `add.ts`, deleting the stale single-path copy. No behavior change in dev; bundled installs now find the scripts. Changeset included (patch).

## Testing

- `tests/commands/hooks.test.ts` + `tests/hooks/hooks-cli-integration.test.ts`: 36/36 pass.
- Manually reproduced the failure on a mise install of 5.0.0 and confirmed the scripts copied from `dist/hooks/` install and run clean (the workaround this fix obsoletes).

### DIFF
--- a/.changeset/hooks-add-bundled-source-dir.md
+++ b/.changeset/hooks-add-bundled-source-dir.md
@@ -1,0 +1,5 @@
+---
+'@harness-engineering/cli': patch
+---
+
+`harness hooks add` now resolves hook scripts from the bundled `dist/hooks/` layout. Previously it only probed the dev layout (`src/hooks/`), so every invocation on a published npm/mise install failed with "Hook scripts not found". It now shares `resolveHookSourceDir()` with `hooks init`, which already handled both layouts.
--- a/packages/cli/src/commands/hooks/add.ts
+++ b/packages/cli/src/commands/hooks/add.ts
@@ -1,22 +1,14 @@
 import { Command } from 'commander';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
-import { fileURLToPath } from 'node:url';
 import { HOOK_SCRIPTS } from '../../hooks/profiles';
 import { supportFilesFor } from '../../hooks/support-files';
 import { logger } from '../../output/logger';
-
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
+import { resolveHookSourceDir } from './init';
 
 const ALIASES: Record<string, string[]> = {
   sentinel: ['sentinel-pre', 'sentinel-post'],
 };
-
-function hookSourceDir(): string {
-  const d = path.resolve(__dirname, '..', '..', 'hooks');
-  if (fs.existsSync(d)) return d;
-  throw new Error(`Hook scripts not found: ${d}`);
-}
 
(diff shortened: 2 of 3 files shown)


chadjw (approved):

chadjw:
👍

PR #954, opened 2026-07-21
test(orchestrator): POSIX-guard codex fakeCodex-spawn tests (green Windows CI)

chadjw:
## Summary

Greens `build-and-test (windows-latest)`, which has been red on `main` (and on every PR — non-blocking, `main` has no required checks) since the codex backend landed.

## Root cause

Three `CodexBackend` tests (`runTurn streams JSONL…`, `…non-zero exit`, `…wall-clock cap`) spawn a `fakeCodex` stand-in written as a `#!/bin/sh` script. `CodexBackend.runTurn` launches `this.command` with a **shell-less** `child_process.spawn`. On Windows that can execute only a real `.exe`:
- a shebang script is not honored (no `/bin/sh`, no shebang support), and
- a `.cmd`/`.bat` throws `EINVAL` without `shell: true` (Node's CVE-2024-27980 fix, present in the pinned Node 22).

So the POSIX shell-script fixture cannot be spawned on Windows by construction. A cross-platform fixture isn't achievable — a test can't ship a compiled `.exe`, and the backend intentionally spawns shell-less (changing that to `shell: true` would be a production behavior/quoting change for a non-Windows feature).

## Fix

Guard exactly those three fixture-spawn tests with `itPosix = it.skipIf(process.platform === 'win32')`, with a documenting comment. The other tests (`name`, no-model, `getModel`, `healthCheck`) don't depend on a script stand-in and continue to run on all platforms. Matches the repo's existing bash-hook e2e `skipIf(win32)` convention. Test-only — no runtime change (empty no-release changeset).

## Verification

- macOS: **13/13 pass** (unchanged).
- Simulated `win32` (predicate forced true): **10 pass, 3 skip, 0 fail** → suite green.
- Full pre-commit (arch gate) + pre-push gauntlet green.

### DIFF
--- a/.changeset/codex-test-windows-skip-no-release.md
+++ b/.changeset/codex-test-windows-skip-no-release.md
@@ -1,0 +1,5 @@
+---
+
+---
+
+Test-only: POSIX-guard the three `CodexBackend` tests that spawn a `#!/bin/sh` `fakeCodex` stand-in (`it.skipIf(win32)`). Node's shell-less `child_process.spawn` can launch only a real `.exe` on Windows — not a shebang script or a `.cmd` (EINVAL) — so these fixture-spawn tests are POSIX-only, matching the repo's existing bash-hook e2e `skipIf(win32)` convention. Greens `build-and-test (windows-latest)`. No runtime change.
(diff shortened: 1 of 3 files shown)


chadjw:
Superseded by #948 — the POSIX-guard for the fakeCodex-spawn tests is folded directly into #948 (which also adds a fourth fixture-spawn test, the `--sandbox workspace-write` argv check, that needs the same guard). Closing to avoid a codex.test.ts conflict.

PR #986, opened 2026-07-31
fix(review): stop the security floor tier emitting fabricated criticals (prose SQL, test fixtures, comment bodies)

### LINKED ISSUE #984: review-ci floor tier emits a fabricated critical: file-length finding merged with SQL-injection metadata (CWE-89 on a help string)
## Summary

`harness review-ci` (floor-only tier, no `ANTHROPIC_API_KEY`) emitted a **single finding whose metadata and body come from two different findings**. The result is a confidently-labelled `critical` / `security` / CWE-89 SQL-injection finding pointing at a line with no SQL in it, which **blocks the PR**.

Observed on PR #983, run [30596263261](https://github.com/Intense-Visions/harness-engineering/actions/runs/30596263261).

## The finding, verbatim

```json
{
  "title": "[architecture, security] Potential SQL injection via string concatenation",
  "domain": "security",
  "severity": "critical",
  "cweId": "CWE-89",
  "owaspCategory": "A03:2021 Injection",
  "confidence": "high",
  "trustScore": 56,
  "validatedBy": "heuristic",
  "subagent": "typescript-strict",
  "rationale": "Past 400 lines a single TypeScript file typically encodes more than one responsibility. Adding to it compounds the burden on future readers and reviewers.",
  "suggestion": "Identify two responsibilities and split the larger one into a new module. A new file with a clear name reads better than a longer file with vague boundaries.",
  "evidence": [
    "File has 442 lines (threshold: 300)",
    "File length: 442 lines",
    "Line 409: 'never create a ticket for a row lacking an externalId (report the skip instead) — ' +"
  ],
  "remediation": "Use parameterized queries or a query builder (e.g., Knex, Prisma) instead of string concatenation."
}
```

## Why it's wrong

The `rationale`, `suggestion`, and first two `evidence` entries are a **file-length** finding (442 lines vs a 300 threshold). The `title`, `domain`, `severity`, `cweId`, `owaspCategory`, and `remediation` are a **SQL-injection** finding. They have been merged into one object.

The cited line is a `commander` option description:

```ts
.option(
  '--no-create',
  'never create a ticket for a row lacking an externalId (report the skip instead) — ' +
    'an unattended job must not invent issues'
)
```

`packages/cli/src/commands/roadmap/sync.ts` contains **no SQL, no database access, and no query construction of any kind** — it is a CLI command definition. The only thing it shares with the SQL-injection heuristic is a `+` between two string literals.

## Why this matters more than a cosmetic mislabel

1. **It blocks merges on a fabricated critical.** `severity: critical` + `domain: security` is the strongest signal the tool can emit, spent on a help string.
2. **It is unfalsifiable from the report alone.** A reviewer reading "critical SQL injection, CWE-89, confidence high" has no way to tell it's spurious without opening the file. `trustScore: 56` and `validatedBy: "heuristic"` are the only hints, and they are buried under a `confidence: "high"` that contradicts them.
3. **It teaches people to ignore security findings.** One fabricated critical does more damage to the gate's credibility than ten missed real ones, because the next genuine CWE-89 gets waved through as "probably that string-concat false positive again."
4. **It fires precisely where verification is weakest** — the floor-only path, with `skipReason: "LLM tier skipped — secret ANTHROPIC_API_KEY not set (floor-only)"`. The tier least able to sanity-check itself is the one emitting `critical`.

## Suspected cause

A finding-aggregation step in the floor tier is merging or aliasing objects — most likely reusing/mutating a shared finding object, or zipping heuristic hits to metadata by index when the two lists differ in length. Both findings appear to originate from the `typescript-strict` subagent on the same file, which is consistent with an index-misalignment inside one subagent's result set rather than a cross-subagent merge.

## Suggested fixes

- [ ] Find and fix the aggregation bug (deep-copy findings, or key metadata to a finding id instead of position). A regression test with two heuristic hits on one file, asserting each keeps its own title/severity/CWE, would lock it.
- [ ] **Add an invariant check before emitting**: a finding whose `domain`/`cweId` implies a vulnerability class must carry evidence consistent with that class. A CWE-89 finding whose only evidence is "file length: 442 lines" should be dropped or downgraded, not published as `critical`.
- [ ] **Reconcile `confidence` with `validatedBy`/`trustScore`.** `validatedBy: "heuristic"` with `trustScore: 56` should not surface as `confidence: "high"`. Heuristic-only findings arguably shouldn't be able to reach `critical` at all without the LLM tier confirming them.
- [ ] Consider whether the floor tier should be allowed to **block** on `security`/`critical` at all when the LLM tier was skipped — or whether it should report and let a human adjudicate.

## Note on the legitimate signal inside it

The file-length half is a **real** finding — the file is 441 lines against a 300-line threshold, and splitting it is reasonable. That part will be addressed in PR #983. It is only the SQL-injection framing that is fabricated. Worth noting because a naive "false positive → suppress it" response would also suppress the true finding underneath.

bstevenski-capillary:
Closes #984.

## What this fixes

`review-ci`'s floor tier was emitting **blocking `critical` security findings for strings that cannot reach any sink**. Two distinct false-positive classes, both observed on real PRs:

1. **Prose that uses a SQL keyword as an English word.** A `commander` help string — `'never create a ticket for a row lacking an externalId … ' +` — was reported as `critical` CWE-89 SQL injection, purely because "create" whole-word-matches `CREATE` and the literal is followed by `+`. Whole-word matching already spared inflected forms ("created"/"updated"); the bare stem was the hole.
2. **Fixtures and documentation.** A test that proves a detector *fires* must contain the vulnerable shape as data, and a JSDoc documenting the shape a rule detects necessarily contains it too. Both were scanned, so **any PR touching a security test or its docs self-flagged** — including this file's own JSDoc, reported as a critical CWE-89.

## The fix, per detector

- **SQL:** a statement keyword alone is not evidence of SQL. Real queries pair one with a **structural companion** (`SELECT … FROM`, `INSERT INTO`, `UPDATE … SET`, `DELETE FROM`, `… TABLE/JOIN/VALUES`); prose essentially never does. Requiring both inside the same literal keeps every genuine injection shape firing and drops the prose class.
- **eval / shell / SQL:** already gated on `isCodeFile`; now also skip **comment lines** (`//`, `/*`, `*` JSDoc bodies).
- **Secrets — the fourth detector, which had neither guard:** now skips test files and comment bodies.

### One deliberate asymmetry

The secrets detector still does **not** gate on `isCodeFile`. A credential in a dotenv or shell file is a real leak, whereas `eval(`, backtick-`exec(` and SQL concatenation are code constructs. Restricting it would lose coverage, not noise.

**Scope caveat, measured rather than assumed:** `SECRET_PATTERNS` key off an assignment shape (`<name> = "<value>"`), so a dotenv/shell line **is** matched while a YAML mapping (`apiKey: "..."`) is **not**. I wrote a comment claiming YAML coverage, the test written to prove it failed, and I corrected the comment rather than the expectation. The wider file scope is real but narrower than "all config files" — that YAML gap is pre-existing and left for its own issue rather than widened inside a false-positive PR.

## Tests — both directions

35 passing in `security-agent.test.ts`, including four new cases pinning that the guards **cannot over-suppress**:

- a fake key in a test fixture → no finding
- an example key in a JSDoc body → no finding
- a real hardcoded key in `src/` → **still `critical`**
- a real key in a non-code file (`.env.example`) → **still fires**

## Why this matters beyond the noise

The floor tier runs when `ANTHROPIC_API_KEY` is unset, so the tier least able to sanity-check itself was publishing `critical` / `confidence: high` findings carrying `validatedBy: heuristic` and `trustScore: 56`. One fabricated critical costs more than ten missed real ones, because the next genuine CWE-89 gets waved through as "probably that string-concat thing again."

#984 also suggests two structural hardenings **not** done here: an invariant that a finding claiming a vulnerability class must carry evidence consistent with that class, and reconciling `confidence` with `validatedBy`/`trustScore`. Worth doing separately; this PR fixes the detectors.

## Provenance

Both commits were originally authored on the `roadmap sync` branch (#983) while investigating why that PR's `required-review` gate was red. Mixing a review-engine change into a roadmap-CLI PR was a packaging mistake, so they were moved here and #983 was force-pushed back to roadmap-only scope (no reviews had been submitted, so nothing was invalidated). #983 is unaffected and independently green apart from the gate this PR fixes.

### DIFF
--- a/.changeset/security-agent-sql-prose-fp.md
+++ b/.changeset/security-agent-sql-prose-fp.md
@@ -1,0 +1,34 @@
+---
+'@harness-engineering/core': patch
+---
+
+fix(review): stop the SQL-injection heuristic firing on prose that uses a SQL keyword as a word
+
+The security review agent's CWE-89 heuristic reported a **critical** SQL-injection
+finding for a `commander` help string:
+
+```
+'never create a ticket for a row lacking an externalId (report the skip instead) — ' +
+```
+
+There is no SQL there. The pattern matches case-insensitively, so the ordinary
+English word "create" whole-word-matched the `CREATE` keyword, and the literal is
+followed by `+`. Whole-word matching already spared inflected forms
+("created"/"updated"); the bare stem was the hole. This is the same class of
+false positive as issue #657 — whose string-boundary fix was necessary but not
+sufficient — and it produced a `critical` finding with `trustScore: 49` on a PR
+containing no database code at all.
+
+A statement keyword alone is not evidence of SQL. Real queries pair one with a
+**structural companion** token (`SELECT … FROM`, `INSERT INTO`, `UPDATE … SET`,
+`DELETE FROM`, `CREATE`/`ALTER`/`DROP TABLE`, `… JOIN`, `… VALUES`); prose
+essentially never does. Both the concatenation and template-literal alternatives
+now require a keyword **and** a companion inside the same literal.
+
+Every genuine injection shape still fires (verified by parametrized tests for
+`INSERT INTO`, `UPDATE … SET`, `DELETE FROM`, and an interpolated
+`SELECT … JOIN … WHERE`), and the prose class is gone — including a regression
+test pinned to the verbatim line that blocked the PR. The pre-existing
+nested-quote blind spot (`"INSERT INTO t VALUES ('" + name + "')"`, which the
+`[^"']` class cannot span) is now documented as a known limitation rather than
+left implicit; it was not introduced by this change.
--- a/.changeset/security-agent-sql-prose-fp.md
+++ b/.changeset/security-agent-sql-prose-fp.md
@@ -2,33 +2,46 @@
 '@harness-engineering/core': patch
 ---
 
-fix(review): stop the SQL-injection heuristic firing on prose that uses a SQL keyword as a word
+fix(review): stop the security floor tier emitting fabricated criticals (#984)
 
-The security review agent's CWE-89 heuristic reported a **critical** SQL-injection
-finding for a `commander` help string:
+The no-LLM security floor was reporting blocking `critical` findings for strings
+that cannot reach any sink. Three false-positive classes, all observed on real PRs:
 
-```
-'never create a ticket for a row lacking an externalId (report the skip instead) — ' +
-```
+- **Prose using a SQL keyword as an English word.** A `commander` help string —
+  `'never create a ticket for a row lacking an externalId … ' +` — was reported
+  as critical CWE-89 because "create" whole-word-matches `CREATE` and the
+  literal is followed by `+`. Same class as issue #657, whose string-boundary
+  fix was necessary but not sufficient.
+- **Test fixtures.** A test proving a detector fires must contain the vulnerable
+  shape as data, so any PR touching a security test self-flagged.
+- **Comment bodies.** A JSDoc documenting the shape a rule detects necessarily
+  contains it; the rule's own JSDoc was reported as a critical CWE-89.
 
-There is no SQL there. The pattern matches case-insensitively, so the ordinary
-English word "create" whole-word-matched the `CREATE` keyword, and the literal is
-followed by `+`. Whole-word matching already spared inflected forms
-("created"/"updated"); the bare stem was the hole. This is the same class of
-false positive as issue #657 — whose string-boundary fix was necessary but not
-sufficient — and it produced a `critical` finding with `trustScore: 49` on a PR
-containing no database code at all.
+The fixes:
 
-A statement keyword alone is not evidence of SQL. Real queries pair one with a
-**structural companion** token (`SELECT … FROM`, `INSERT INTO`, `UPDATE … SET`,
-`DELETE FROM`, `CREATE`/`ALTER`/`DROP TABLE`, `… JOIN`, `… VALUES`); prose
-essentially never does. Both the concatenation and template-literal alternatives
-now require a keyword **and** a companion inside the same literal.
+- **SQL: ordered statement shapes.** A statement keyword alone is not evidence
+  of SQL. The pattern now requires an ordered shape (`SELECT … FROM`,
+  `INSERT INTO`, `UPDATE … SET`, `DELETE FROM`, `CREATE/ALTER/DROP TABLE`,
+  `UNION SELECT`) inside a concatenated string literal or an interpolated
+  template literal. The vocabulary deliberately mirrors `SQL_QUERY_SHAPE` in
+  `finding-integrity.ts`, so nothing the floor emits is downgraded by the
+  Phase 5.75 integrity invariant (#989) — one definition of "looks like SQL",
+  not two. The template alternative does not require a closing backtick, so the
+  opening line of a multi-line template query still fires.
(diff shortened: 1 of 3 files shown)


chadjw (dismissed):

chadjw:
## Code review (harness-code-review, deep tier) + fixes applied

**Assessment: Request Changes → fixed in d9f92c125** (rebased onto main; the branch was 9 commits behind, which also explains the earlier doc-drift `harness` check failure on files this PR never touched).

Five review agents (bug, security, adversarial, compliance/architecture, typescript-strict) ran against the diff; every finding below was verified by executing the actual regexes, old vs new.

### Important findings (all fixed)
1. **SQL recall regressions**: the keyword+companion rewrite silently dropped split-literal queries (`"SELECT " + cols + " FROM users"`), multi-line concatenated queries, and multi-line template literals (the new template alternative required a closing backtick on the same line — the old one didn't). The JSDoc/changeset claim "every genuine injection shape still fires" was false.
2. **`WHERE` sat in both the template-keyword and companion lists**, so one prose word satisfied both lookaheads — ```logger.info(`this is where ${x} lives`)``` still fired as critical CWE-89.
3. **#989 semantic divergence**: the unordered keyword×companion vocabulary fired on shapes (`SELECT … WHERE` with no `FROM`) that main's `finding-integrity` Phase 5.75 immediately downgrades — two definitions of "looks like SQL".
4. **`isCommentLine` under-scanned**: `/**/ eval(x)`, `*/ code`, and generator members (`*run() { … }`) were skipped by all four detectors, while its JSDoc claimed it "can only over-scan, never under-scan". A uniform 4-char prefix could silence the whole floor.
5. **Secrets guards applied JS conventions to every file type**: `.env.test.local` (live staging creds by convention) and a key in a Markdown `*` bullet became invisible.

### The fix
- SQL pattern now requires **ordered statement shapes** (`SELECT…FROM`, `INSERT INTO`, `UPDATE…SET`, `DELETE FROM`, `CREATE/ALTER/DROP TABLE…`, `UNION SELECT`) mirroring `finding-integrity`'s `SQL_QUERY_SHAPE` — one vocabulary, and a new cross-layer test asserts detector output survives `enforceFindingIntegrity` undowngraded. Closing backtick no longer required (multi-line template queries fire again).
- `isCommentLine` skips **comment-only** lines; trailing-`//` strip is URL-safe (`(?<!:)//` — the old `indexOf('//')` truncated at `https://`, a pre-existing hole).
- Test-file markers + comment syntax now apply **only to code-extension files**; secrets keep the wider file scope.
- `isCodeFile` → `isScannableSourceFile` (sibling `bug-agent.ts` has an `isCodeFile` with different semantics).
- Split-literal / multi-line-concat misses are **documented and pinned as must-miss tests** rather than claimed away — loosening the shape to line level would resurrect the concatenated-help-string prose FP class this PR exists to kill.

51 tests (was 35), all green; `tsc --noEmit` clean; full `tests/review/` suite (482) green against merged main.

### Deliberate trade-off left in place
A real credential pasted into a `*.test.ts` is still skipped by the secrets detector — there's no mechanical way to tell it from the fake keys security tests must contain. Now documented in the detector JSDoc as LLM-tier territory rather than implied to be safe.

PR #1008, opened 2026-08-02
feat(skills): harness-repo-hygiene — safe fleet-wide branch/worktree pruning

bstevenski-capillary:
Adds `harness-repo-hygiene`, a **rigid** 6-phase pipeline for pruning many branches and worktrees at once: SYNC → CLASSIFY → AUDIT → PRUNE → TEARDOWN → REPORT.

## Why a new skill rather than extending `harness-git-workflow`

`harness-git-workflow` owns the **single-workstream lifecycle** — create worktree, install, verify baseline, then finish or discard one branch. It has no bulk prune, no fetch-first step, and no cross-branch unpushed audit; its "Discard" path handles one branch with a confirmation. Different scope, so the two compose rather than overlap. `related_skills` links them.

The narrow `[gone]`-branch cleanup that third-party tools (e.g. the `clean_gone` slash command) perform is **Phase 4 step 2** here — a stage in the pipeline, not a competing tool. It passes through the AUDIT gate first, since a deleted upstream is precisely when unpushed commits become unrecoverable.

## Every safety rule comes from an observed failure

Derived from a real 68-branch / 23-worktree sweep, not invented:

| Rule | Failure it prevents |
| --- | --- |
| SYNC is Phase 1 | A sweep reported "cleanup complete" with every ref 3–613 commits behind |
| CLASSIFY checks containment in **open PR branches**, not just `main` | 3 branches read as unmerged against `main` were fully contained in an open PR |
| AUDIT requires `AHEAD=0`; `no-remote` escalates **individually** | A `backup/` branch held 7 commits present on no remote |
| TEARDOWN documents the half-deleted worktree state | An interrupted removal left 687 files deleted and the tree still registered |
| PRUNE mandates file/array iteration | In `zsh`, `for b in $LIST` does not word-split — the loop reports success having deleted nothing |
| REPORT returns ref state, not a delete count | "31 deleted" is compatible with every surviving ref being stale |

## Testing

4 scenarios per `harness-skill-authoring` Phase 5B, exercising both Red Flags, a Rationalization, and both Gates. All 8 required sections verified present (rigid ⇒ Gates + Escalation included). Identical copies across claude-code / cursor / codex / gemini-cli.

## Two gaps found while authoring

1. **`harness skill validate` cannot validate a new skill.** It scans the installed CLI bundle rather than the working tree. I confirmed this with a deliberate break — removing the `## Rationalizations to Reject` heading produced **no error**. Its silence is an abstention, not a pass. Required sections here were verified by hand.

2. **`docs/reference/cli-commands.md` and `mcp-tools.md` were already drifted** — `pnpm run generate-docs` updated them alongside the catalog. Included here since the catalog regeneration required the same command; happy to split them out.

Catalog goes 766 → 767.

### DIFF
--- a/agents/skills/claude-code/harness-repo-hygiene/SKILL.md
+++ b/agents/skills/claude-code/harness-repo-hygiene/SKILL.md
@@ -1,0 +1,299 @@
+# Harness Repo Hygiene
+
+> Fleet-wide branch and worktree pruning. Sync first, classify every ref against PR state and containment, prove nothing unpushed will be lost, then prune — and report the resulting ref state, not a delete count.
+
+## When to Use
+
+- When a repo has accumulated dozens of local branches and worktrees and you want the dead ones gone
+- When agent-driven work has left `worktree-agent-*` / `.claude/worktrees/*` debris behind
+- When cross-worktree tool caching (turbo, nx, bazel) is contaminating a build or bisect and you need a single-tree checkout
+- When a reviewer requires branches to be up to date and you are preparing a queue for review
+- NOT for creating a worktree or finishing a single branch — that is `harness-git-workflow`
+- NOT for deleting remote branches; this skill only prunes local refs and worktrees
+- NOT as a pre-commit or CI step — this is an interactive maintenance sweep with human confirmation points
+
+## Process
+
+### Iron Law
+
+**Never delete a ref you have not proven exists somewhere else.**
+
+Every deletion in this skill is justified by evidence gathered in Phase 2 and verified in Phase 3 — a merged PR, containment in `main`, containment in an open PR branch, or an explicit human decision. A branch whose commits exist nowhere but this disk is not prunable, no matter how old or how oddly named. "Looks abandoned" is not evidence.
+
+---
+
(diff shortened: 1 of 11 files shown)


chadjw:
Review pass: the branch's docs regeneration was run with a stale CLI build and **deleted unrelated, live documentation** — the `--allow-regress`/`--reason` flags for `harness check-arch` (docs/reference/cli-commands.md) and the entire `edit_file` MCP tool section (docs/reference/mcp-tools.md). Both features exist on main.

Pushed f49d03f restoring those two files to merge-base state — this PR's only legitimate reference-docs change is the skills-catalog entry (766→767), which is kept. If you regenerate docs again on this branch, use the locally built CLI (`node packages/cli/dist/bin/harness.js generate-docs`), not the global install.

Non-blocking suggestion: Phase 2's per-branch `gh pr list --head "$branch"` is one GraphQL call per branch; a single batched `gh pr list --state all --json headRefName,state,number` covers the whole sweep. Skill content otherwise looks solid — 4 platform copies are byte-identical and the irreversibility gates (AHEAD=0, per-branch no-remote escalation) are well constructed.

Issue #1091, opened 2026-08-05
flaky: otlp-http exporter test asserts receiver.received is empty while a retry may still be in flight


ahhrealmonster:
`packages/core/src/telemetry/exporter/otlp-http.test.ts` →
**"drops the batch after 3 failed attempts and warns once"** fails
intermittently on unchanged code. Two consecutive runs, same commit, no edits
between them:

```
$ npx vitest run src/telemetry/exporter/otlp-http.test.ts
 Tests  6 passed (6)                      # 523ms

$ npx vitest run src/telemetry/exporter/otlp-http.test.ts
 Tests  1 failed | 5 passed (6)           # 2.47s
```

The failing assertion is the last line, `expect(receiver.received).toHaveLength(0)`
(around line 149).

## Likely cause

The test mixes fake timers with a **real** HTTP receiver:

```ts
vi.useFakeTimers({ shouldAdvanceTime: true, shouldClearNativeTimers: false });
// ...advance 1s/2s/4s twice to burn the retry backoff...
vi.useRealTimers();

await vi.waitFor(() => expect(warn).toHaveBeenCalledTimes(1), { timeout: 2000 });
expect(receiver.received).toHaveLength(0);
```

Timer advancement is synchronous but the retry POSTs are not — they are in flight
against a real socket. Once `useRealTimers()` restores the clock, an
already-dispatched retry can still land on the receiver, so `received` picks up an
entry and the length-0 assertion fails. The 523ms-vs-2.47s duration split matches
that reading: the slow run is the one where a request was still outstanding. The
2.47s run is also suspiciously close to the `waitFor` 2000ms timeout, so the
`warn` assertion may be racing too.

The assertion is checking the right property — a dropped batch must not reach the
collector — it just cannot be evaluated at a moment when delivery is still
possible.

## Suggested fix

Make the "no delivery happened" claim deterministic rather than time-bounded:

- await quiescence explicitly (drain the exporter's in-flight promises, or expose a
  test seam that resolves when the retry chain has terminated) before asserting on
  `receiver.received`; or
- assert on the receiver's **request count observed by the exporter** (attempts
  made, all 503) rather than on payloads accepted, so a late arrival cannot change
  the answer; or
- serve the 503s from a fake transport instead of a real socket in this test, and
  keep the real-receiver coverage in the happy-path tests where delivery is the
  thing under test.

Found while running the core suite for an unrelated change
(#1084 / the check-security scan-surface fix); my diff does not touch
[...]

PR #1093, opened 2026-08-05
fix(security): check-security scans no .mjs/.cjs — and a zero-file scan prints as a pass

### LINKED ISSUE #1084: check-security scans no .mjs — an ESM-only repo gets a green gate that read ~none of its code, and a zero denominator prints as a pass
## Summary

`harness check-security` never scans `.mjs` files. Its file glob is

```
**/*.{ts,tsx,js,jsx,go,py,java,rb}
```

(`dist/chunk-VMAJGAFL.js` in `@harness-engineering/cli@10.2.0`; the sibling glob
adds `sh` and still omits `mjs`). Any repo whose executable surface is ESM
`.mjs` therefore gets a **green security gate that scanned essentially none of
its own code** — and green-because-empty is indistinguishable from green-because-clean
in the current output.

## Reproduction

Same two secrets, three extensions, same directory:

```js
// scripts/_probe.ts   AND   scripts/_probe.mjs
export const k = 'AKIAIOSFODNN7EXAMPLE'; // SEC-SEC-001
export const api_key = 'abcdef1234567890'; // SEC-SEC-002
```

```
$ harness check-security --severity error --json
issues: 4
   scripts/_probe.ts:1
   scripts/_probe.ts:2
   scripts/_probe.py:1
   scripts/_probe.py:2
```

The byte-identical `.mjs` file is absent from the findings. Reproduced on
`@harness-engineering/cli@10.2.0` and `@4.1.0`, so this is long-standing rather
than a recent regression.

## Why it bites

In the repo where I hit this, `harness check-security --severity error` is a
**blocking** CI step, and the code surface is:

| extension | tracked files | scanned? |
| --- | --- | --- |
| `.mjs` | **144** | no |
| `.ts` | 32 (mostly test fixtures and two skill runtimes) | yes |
| `.js` | 6 | yes |
| `.py` | 1 | yes |

So the gate has been passing on roughly the 39 files that are *not* the
project's actual implementation, while its 144 real source files went unread.
That state produced a clean security ledger with `securityScore: 100,
totalFindings: 0` — a score that reads as an assurance and is really an
abstention. The repo had recently completed a Python-to-Node port, which is
precisely when the coverage silently went to near-zero and nothing said so.

## Asks

1. Add `mjs` (and `cjs`, `mts`, `cts` for the same reason) to the scan glob.
2. Report the denominator: print files-scanned alongside findings, and treat a
   scan that matched **zero** files as a distinct non-pass outcome rather than
   `valid: true, issues: []`. A gate that read nothing did not pass; it
   abstained. Same argument for `--changed-only` when the changed set contains no
   scannable file.
3. Consider deriving the glob from the project's configured language/tooling
   (`harness.config.json` already declares `template.language` and
   `tooling.packageManager`) so an ESM-only project cannot end up with an empty
   scan set by default.

Item 2 is the one that would have surfaced this on day one — the extension list
will always lag some ecosystem, but a printed zero denominator cannot hide.

bstevenski-capillary:
Closes #1084.

## The bug

`check-security` globbed `**/*.{ts,tsx,js,jsx,go,py,java,rb}`. No `.mjs`, no
`.cjs`. An ESM-only Node project gets a scan that matches none of its source, and
`valid: true, issues: []` — which is byte-identical to the output of a genuinely
clean run. Green because empty.

Same two secrets, same directory, three extensions:

```js
export const k = 'AKIAIOSFODNN7EXAMPLE'; // SEC-SEC-001
export const api_key = 'abcdef1234567890'; // SEC-SEC-002
```

```
$ harness check-security --severity error --json
issues: 4
   scripts/_probe.ts:1     scripts/_probe.ts:2
   scripts/_probe.py:1     scripts/_probe.py:2
```

The `.mjs` twin is absent. Reproduced on `@10.2.0` and `@4.1.0` — long-standing,
not a recent regression.

Where I hit it, `check-security --severity error` is a **blocking** CI step and the
repo had just finished a Python→Node port:

| extension | tracked files | scanned before |
| --- | --- | --- |
| `.mjs` | **144** | no |
| `.ts` | 32 (mostly fixtures + 2 vendored runtimes) | yes |
| `.js` / `.py` | 7 | yes |

So the gate was reading the ~39 files that aren't the implementation and passing on
that basis, with a security ledger recording `securityScore: 100, totalFindings: 0`.

## Changes

**One scan surface.** The glob literal was copy-pasted in three places and had
already drifted — the CI check-orchestrator's copy also omitted `java`/`rb`. It now
lives in `core/src/security/scan-targets.ts` as `SECURITY_SCAN_GLOB` /
`SECURITY_SCAN_EXTENSIONS` / `SECURITY_SCAN_DEFAULT_IGNORE`, consumed by
`check-security`, the orchestrator, and the dashboard gatherer. `mts`/`cts` added
alongside `mjs`/`cjs`.

**The denominator is now visible** — this is the part that would have caught the bug
on day one, since an extension list will always lag some ecosystem but a printed
zero cannot hide:

- text: appends `N file(s) scanned, M rule(s) applied`
- JSON: adds `scannedNothing` and `stats` next to the verdict
- a zero-file scan emits an explicit `ABSTAINED: 0 files scanned — nothing was
  verified` issue instead of presenting as clean
- new `--fail-on-empty` makes that abstention blocking

**Zero-file scans stay non-blocking by default.** A repo may legitimately have no
scannable source, and reddening every such repo on upgrade would be a worse failure
than the one being fixed. Gates that want the strict reading opt in — I'd suggest
`ci.yml` adopts `--fail-on-empty` in a follow-up, but that is your call, not a
drive-by.

## Verification

- 8 new tests (15 in the file). **Red-test checked:** removing `mjs`/`cjs` from the
  extension list turns 4 of them red, so they detect the actual bug rather than
  restating the fix.
- New fixtures: `security-esm` (ESM/CJS project with a real finding) and
  `security-no-source` (markdown only → must abstain).
- The extension list is pinned by an explicit assertion, so a future edit can't
  silently drop one without also re-running the behavioural tests.
- `turbo typecheck`, `lint`, `test:coverage --affected`, prettier, changeset check,
  `generate-docs --check` all green. CLI 5,365 tests and core 3,859 pass.

## The fix immediately found real bugs — in this repo

With `.mjs` finally in the glob, `harness ci check` reported **3 error-severity
`SEC-INJ-003` findings in harness's own scripts**. They were always there; nothing
was reading those files. Suppressing them with `harness-ignore` would defeat the fix
being shipped, so each is fixed at source — all three become argv arrays with no
shell:

| file | what it concatenated |
| --- | --- |
| `scripts/generate-docs.mjs` | `execSync(\`node ${promptArgs.join(' ')}\`)` → `execFileSync(process.execPath, promptArgs)` |
| `scripts/benchmark-check.mjs` | `outputFile` (package name + absolute path) interpolated into the command |
| `docs/guides/recipes/ci-check-script.mjs` | **`HARNESS_FAIL_ON` / `HARNESS_SKIP` / `HARNESS_REPORT` env vars into a `sh -c` string with a `>` redirect** |

The third is the one that matters: it is a recipe people copy into their own CI, so a
concatenated command there hands whoever controls those variables arbitrary execution
on the runner. It now builds an argv array and writes the report with `writeFileSync`,
capturing stdout in-process — the shell redirect was the only thing that needed a
shell.

After: `harness ci check --skip arch` exits 0 (was 1), and `check-security` reports
**1797 files scanned, 0 errors** — a denominator that is printed rather than assumed.

## Notes for the reviewer

**No baseline or catalog churn.** An earlier revision of this branch had to accept an
arch-baseline regression and a `skills-catalog.md` count bump, both of which were
pre-existing drift on main rather than this change. Main has since refreshed its own
baseline, so after merging main in, this branch's diff against those generated files
is **empty**. The only generated file still in the diff is `docs/roadmap.md`, which
the mandatory pre-commit hook regenerates from `docs/roadmap.d/` — the sanctioned
freshness path per AGENTS.md, and unrelated to this fix in content.

## Also filed while verifying this

- **#1091** — `otlp-http.test.ts` "drops the batch after 3 failed attempts" is
  flaky: it asserts `receiver.received` is empty while a retry may still be in
  flight against a real socket. Two consecutive runs on unchanged code, one pass one
  fail.
- **#1094** — `cli#test:coverage` and `orchestrator#test:coverage` each fail
  intermittently under the pre-push `--concurrency=2` run while passing standalone,
  which blocked this push three times. The compound parallel-load flake the hook's
  own comment predicts; filed with the retry evidence.


ahhrealmonster:
Status: **8 of 9 checks green**, including `harness` (the injection findings this fix exposed are now fixed at source) and `build-and-test (macos)` — the earlier macOS red was the load flake now tracked as #1094.

The remaining red is `build-and-test (windows-latest, 22)`, which is **not caused by this PR**. Windows is already red on `main` (in `core#test`), and because turbo stops at the first failed task, `cli#test` never runs there — so two Windows-only failures from recently merged features were masked and only became visible here:

- `tests/commands/outcome-eval-ci.test.ts:69` — asserts a POSIX separator literally (`/repo/...` vs `\repo\...`), from #1077
- `src/commands/roadmap/install-hook.test.ts:139` — `expected +0 not to be +0`, from #1078

Filed as **#1096** with the masking mechanism. This PR's only CLI change is `check-security`; it can't fix those and shouldn't be read as having broken Windows.

chadjw (approved):

PR #1097, opened 2026-08-05
fix(cli): guard two Windows-only test assertions blocking the fleet

chadjw:
## Problem

Every open PR shows `mergeStateStatus=BLOCKED` because `main` itself is red on `build-and-test (windows-latest, 22)`. Two test-only assertions fail on Windows:

1. **`packages/cli/src/commands/roadmap/install-hook.test.ts`** — asserts the pre-commit hook has POSIX exec bits. Since the chmod hotfix (#1092) correctly guards `chmodSync` behind `process.platform !== 'win32'` (git for Windows honors the hook regardless of mode), the exec bits are never set on Windows and the assertion fails. The test wasn't guarded to match.
2. **`packages/cli/tests/commands/outcome-eval-ci.test.ts`** — `resolveSpecPath` returns a native path via `path.join`, but the test hard-coded a POSIX `/repo/...` expectation, so on Windows `\repo\...` !== `/repo/...`.

## Fix

Both are **test-only** corrections — runtime behavior is unchanged:

- Skip the exec-bit assertion on Windows, mirroring the production platform guard.
- Build the expected spec path with `path.join` so it matches the native separator.

## Verification

```
Test Files  2 passed (2)
     Tests  33 passed (33)
```
(Node 22, macOS.)

Once merged, `main` goes green on Windows and unblocks the rest of the open fleet.

### DIFF
--- a/.changeset/windows-test-parity-hotfix.md
+++ b/.changeset/windows-test-parity-hotfix.md
@@ -1,0 +1,14 @@
+---
+'@harness-engineering/cli': patch
+---
+
+Fix two Windows-only CI test failures that block the fleet on `windows-latest`:
+
+- `roadmap install-hook`: the exec-bit assertion now skips on Windows, matching
+  the `process.platform !== 'win32'` guard around `chmodSync` (git for Windows
+  honors the hook regardless of mode).
+- `outcome-eval-ci` `resolveSpecPath`: the expected path is now built with
+  `path.join` instead of a hard-coded POSIX string, so it matches the native
+  separator the implementation returns.
+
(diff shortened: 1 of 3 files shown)


chadjw:
Closing as redundant. PR #1082 (external skill-provider freshness) merged first and independently landed the identical two fixes — guarding the install-hook exec-bit assertion behind `process.platform !== 'win32'` and building the outcome-eval-ci `resolveSpecPath` expectation with `path.join`. `main` is now green on `windows-latest`, so this hotfix adds nothing.

PR #1142, opened 2026-08-06
feat(design-craft): machine-derived award-tier verdict on BENCHMARK output

chadjw:
## What

Adds a machine-derived **award-tier verdict** (`awardBar`) to the design-craft BENCHMARK output. Every `BenchmarkScore` now carries:

```ts
awardBar: {
  verdict: 'cleared' | 'not-cleared' | 'indeterminate';
  dimensions: Record<RadarDimensionName, { score; floor; cleared }>;
  shortfalls: RadarDimensionName[];
  reason?: string;
}
```

computed in **TypeScript, never emitted by the LLM** — the authority-in-TS pattern used by `outcome-eval` / `acceptance-eval`. This replaces free-hand "is this good enough? ~82" judgment with an honest, corpus-calibrated machine signal downstream agents can read directly.

## Why

A consuming project reported `awardBar` was "false on everything." Investigation found two things:

1. **The MarketingPage exemplar corpus already exists here** (9 exemplars, `CRAFT-B009–B017`, ADR 0082). A project that "lacks the corpus" is running a **stale installed `@harness-engineering/cli`** — the fix there is `npm update`, not code.
2. **`awardBar` existed nowhere in the product.** BENCHMARK emitted a 0–100 radar + `min(confidence)` + narrative gaps, but no pass/fail verdict. That gap is what this PR closes.

## How it's computed

- **Per-dimension, not a single overall threshold.** ADR 0082 diagnosed pages scoring 88–94 *overall* while carrying template tells — an equal-weight mean hides the weak axis. Each of the 5 radar dimensions must clear its own floor, so the verdict fails on the exact axis that's weak (ADR 0019).
- **Hybrid exemplar-relative floor:** per dimension, `floor = max(dimensionFloor, round(fraction × median(cited-exemplar references)))`. The corpus defines the bar (no magic number); the config floor stops it eroding; the **median** makes it robust to one weak exemplar.
- **Low confidence forces `indeterminate`** — a high score the model is unsure about must never certify award tier.
- Tunable via `design.craft.benchmark.awardBar` (`dimensionFloor` 80, `fraction` 0.95, `confidenceFloor` medium); omit for defaults.

## Scope note

There is no markdown renderer for BENCHMARK scores anywhere in the codebase, so the verdict is surfaced on the **structured** output (`DesignCraftOutput.scores[].awardBar`) — the machine-consumable surface agents read. Markdown rendering is deferred until a benchmark-score report surface exists (YAGNI).

## Changes

- New pure module `design-craft/phases/award-bar.ts` (`computeAwardBar` + config resolver)
- Schema: `AwardBar` / `AwardVerdict` / `AwardBarDimension` + `awardBar` on `BenchmarkScore`
- Wired through `runBenchmark` / `buildScore` and the MCP pipeline (reads `harness.config.json` defensively; falls back to defaults)
- Config: `design.craft.benchmark.awardBar`
- `CONFIDENCE_RANK` hoisted to the canonical `shared/craft/findings/axes.ts` (reused by benchmark + award-bar)
- ADR 0084; SKILL.md BENCHMARK step across all 4 client copies

## Verification

`tsc` 0 · 73 design-craft tests pass · `eslint src` clean · prettier clean · changeset included. Independent `harness-code-reviewer` pass: **approved**, no critical/important findings.

Spec: `docs/changes/design-craft-award-bar/proposal.md` · ADR: `docs/knowledge/decisions/0084-award-bar-machine-verdict-on-benchmark.md`


chadjw:
Feedback from the consuming side (the `iv-demo` prospect-site fleet that reported "`awardBar` false on everything") — two confirmations and one gap worth weighing before this lands.

**1. The stale-CLI diagnosis is correct.** Confirmed from the consuming environment: the installed `@harness-engineering/cli` is `10.2.0`, and its `dist/` carries neither the MarketingPage exemplar corpus (`CRAFT-B01x`) nor an award-bar module. So "false on everything" really was a stale install, not a scoring bug — `npm update` is the fix, exactly as the "Why" says. The authority-in-TS, per-dimension-floor approach here is the right shape for what we needed.

**2. A live case for the core thesis.** We just ran a hand-audit over this fleet, and it's strong evidence for why a corpus-calibrated machine verdict beats free-hand "~82, good enough": demos that were **structurally clean** (grep-clean tokens, balanced tags, live 200, correct content-types) still failed on rendered inspection. Heuristic/structural pass ≠ award tier — which is precisely the judgment this PR is trying to make honest.

**3. Gap: the verdict has no responsive/mobile dimension.** The 5 radar dimensions are aesthetic (concept-coherence, composition, surface-texture, typography, hierarchy-clarity) — none captures mobile. In our audit, **~10 of 14 built demos carried award-tier-fatal mobile defects** — a nav that `display:none`d to nothing with no hamburger (unreachable on a phone), or 21–75px of horizontal overflow — and **every one of those would clear all five current dimensions**. So a machine `awardBar: cleared` can still certify a page that's broken on the device where it's actually opened.

Recommendation: either (a) add a `responsive` dimension to the corpus + floor computation, or (b) explicitly scope `awardBar` in the schema/docs as **desktop aesthetic clearance, not a mobile/responsiveness gate**, so downstream agents don't read `cleared` as "ship-ready on mobile." Option (b) is cheap and honest today; (a) is the fuller fix once there are mobile exemplars to calibrate against.

Nothing here blocks the PR — the machine verdict is a clear improvement over the status quo. Flagging the mobile dimension so `cleared` doesn't over-promise.

chadjw:
Thanks — the mobile evidence is exactly the kind of thing the machine verdict exists to stop over-promising, so I don't want `cleared` quietly certifying a phone-broken page.

Took **option (b)** now, deferred **(a)**:

- **(a) a `responsive` dimension** can't be done honestly today. It needs a sixth radar axis in the LLM prompt *and* a calibrated `responsive` reference on all 17 exemplars — and there are no mobile-audited exemplars to calibrate against. Adding the axis with invented reference scores would fabricate a bar instead of deriving one from the corpus, which breaks the exemplar-relative principle (D2). So it's deferred, not declined, and named as a gated future increment.
- **(b) explicit scoping** is in as of `14bc00ac7`. `awardBar` is now documented as **desktop aesthetic clearance, not a mobile/responsiveness gate** in three places downstream actually reads:
  - the schema type (`AwardBar` doc comment) — a `SCOPE` paragraph stating a page can clear every dimension while carrying award-tier-fatal mobile defects, and that `cleared` must not be read as "ship-ready on mobile";
  - `SKILL.md` BENCHMARK step (all 4 client copies) — a blockquote scope note;
  - ADR 0084 Consequences — the scope boundary + the `responsive` dimension recorded as a future increment gated on mobile exemplars, citing your ~10/14 audit finding.

Net: `cleared` now honestly means "clears the desktop aesthetic bar," and the responsive gate has a concrete, calibration-gated path when we have exemplars for it. Confirming the stale-CLI point too — 10.2.0 predates both the corpus and the module, so `npm update` is the fix on the consuming side.

Issue #1143, opened 2026-08-06
flaky: subprocess-spawning tests fail intermittently under full-suite parallelism (`pnpm turbo test`)


ahhrealmonster:
Running the whole suite locally (`pnpm turbo test --force`, 10-core machine) fails intermittently, with a **different** test each run. All the observed failures spawn a subprocess or assert on wall-clock behaviour.

Observed failures across 4 full runs:

| Package | Test |
|---|---|
| `signals` | `tests/command-runner.test.ts > runs a command and returns trimmed stdout` |
| `core` | `src/solutions/scan-candidates/git-scan.test.ts > returns only fix: commits within the lookback window` |
| `core` | `tests/review/ci/default-exec-file.test.ts > every failure mode rejects` |
| `core` | `tests/state/event-sourcing/concurrency.test.ts > N processes with distinct writerIds lose zero events` |

Every one of them passes when its file is run alone.

## Not caused by a branch

Verified against a clean detached worktree at `origin/main` (a6fb723f0) with no local changes: 2 of 2 full runs failed, each on a different `core` test. So this is pre-existing on the trunk, not introduced by any in-flight branch.

## Likely mechanism

The failures are consistent with fixed timeouts being exceeded because the machine is saturated, not with a logic defect. For example `packages/signals/src/command-runner.ts:14` spawns via `execFile` with a hard `timeout: 5_000`; under a full-suite run (many vitest workers, each spawning `node`) a bare `node -e` launch can exceed that budget.

Worth noting because it is the same class of defect either way: **a test that fails on machine load teaches people to re-run rather than read the failure**, which is how a real regression gets waved through as "just the flaky one".

## Suggested direction

- Prefer load-invariant assertions over absolute wall-clock budgets. Concrete example from `packages/burn/tests/bin-startup.test.ts`: a startup-cost guard that first used an absolute ceiling (821ms observed against a 700ms limit, purely from parallel load) was rewritten to assert a *ratio* against `node -e ` measured in the same environment — both sides inflate together under load, so it keeps measuring the intended property.
- Where a real subprocess is genuinely needed, raise or remove the fixed timeout rather than tuning it to an idle machine.

## Repro

```bash
pnpm turbo test --force   # repeat a few times; expect a different failure each time
```

chadjw:
[...]

PR #1144, opened 2026-08-06
feat(burn): port the usage-burn HUD to TypeScript as `harness burn`

bstevenski-capillary:
Replaces the standalone [`claude-burn-hud`](https://github.com/bstevenski-capillary/claude-burn-hud) (Python + shell, ~1.4k lines) with an in-repo package, so the gauge and the harness stop drifting apart.

## Two surfaces, split on latency rather than taste

| Surface | What | Why |
|---|---|---|
| `harness burn` | report, `weeks`, `calibrate`, `budget`, `reset-day`, `scan`, `install` | Human-invoked, so the CLI's module graph is affordable. |
| `harness-burn-hud` | `line`, `session-start`, `stop`, `scan` | The statusline repaint and the Stop hook. |

The split is measured, not stylistic. On this machine:

```
statusline.sh (the thing being replaced)   0.11s
node -e     (bare interpreter startup)   0.05s
harness --version (full CLI module graph)  0.85s   ← 8x the repaint budget
```

So `harness-burn-hud` is bundled standalone and imports **nothing** from `@harness-engineering/*`. A test asserts that import graph, because the regression has no symptom other than a terminal that feels slow — nobody bisects that.

## Parity

Both implementations were run over the same **33,305 real transcript records**:

- every shared record **byte-identical**
- with `now` pinned, the summaries differ only in Python's microsecond vs JS millisecond precision and `100.0` vs `100` JSON rendering

## Tests

87 in `packages/burn`, 33 in the CLI command. Every regression test came across from the Python suite, each still tied to a defect that actually shipped:

| Guards against | |
|---|---|
| the Monday-UTC week assumption | understated a 97% week by **~81x** behind a calm green |
| the write race | silently dropped **85%** of the record store while reporting OK |
| repeated usage blocks | inflated every figure **~3.5x** |
| a 3-hour extrapolation | fired CRITICAL at 2% of budget spent |
| a zero/thin denominator | must read as abstention, never as a pass |

## Three runtime-forced changes

Documented at each site:

1. **Locking** — `fcntl.flock` (kernel-released on death) → atomic `mkdir` plus a staleness reclaim, since Node has no `flock` and a crashed holder would otherwise wedge every later scan. Two tests cover what the kernel used to do for free.
2. **Timezones** — `zoneinfo` → `Intl` with two-pass offset resolution, asserted across the 2026-11-01 DST boundary.
3. **Timestamps** — summaries keep Python's `+00:00` form (not Node's `Z`), because `fromisoformat` rejected `Z` before 3.11 and the old HUD may still read these files during cutover.

## Four defects the port introduced and the tests caught

- The binary is emitted as `.mjs` — as `.js` Node was detect-and-reparsing it on **every launch**, which it warns "incurs a performance overhead", on the one path where that matters.
- `line` no longer blocks forever when stdin never closes (surfaced as a 30s test timeout).
- `install` recognises its own previous entry instead of stacking a **duplicate Stop hook on every run** — the bin path is `burn-hud.mjs`, which does not contain the string `harness-burn-hud`.
- The startup guard asserts a **ratio** against bare `node` rather than a wall-clock ceiling. The absolute version failed at 821ms against a 700ms limit purely from parallel test load; under load both sides of a ratio inflate together. Observed: this binary 2.5x bare node, the full CLI 19.2x.

## Cutover

`harness burn install` is additive and reversible: backs up `settings.json`, leaves unrelated hooks alone, reports a statusline take-over rather than silently clobbering, and leaves `~/.claude/hud` on disk so there is a way back. Archiving the standalone repo waits until the TS version has real mileage.

## Gates

Build, typecheck, lint, test, format, coverage ratchet, plugin/barrel checks all green.

- **Arch gate:** every *error-severity* threshold finding was **fixed**, not acknowledged — `renderStatusline`, `renderReport`, `calibrate`, `parseTranscript` and `setResetDay` were each decomposed into named helpers. The gate explicitly refuses to allowance an error-severity breach, which is the right call. The remaining aggregate deltas (the cost of the package existing) are in a per-PR allowance, leaving `baselines.json` byte-identical to the base.
- **Unrelated finding:** the full local suite has a pre-existing flake, reproduced 2/2 on a clean `origin/main` worktree with a different subprocess-timing test failing each run. Filed as #1143 rather than left as folklore.

### DIFF
--- a/.changeset/burn-hud-ts-port.md
+++ b/.changeset/burn-hud-ts-port.md
@@ -1,0 +1,38 @@
+---
+'@harness-engineering/burn': minor
+'@harness-engineering/cli': minor
+---
+
+Ship the usage-burn HUD as `@harness-engineering/burn` + `harness burn`, replacing the
+standalone `claude-burn-hud` Python/shell tool.
+
+The HUD reports Claude Code usage pace from local transcripts: week-anchored spend,
+a baseline-shrunk forecast, per-model family limits, and a `/clear` nudge once the
+checked-out branch has merged. It is a local proxy, never Anthropic's real quota —
+`/usage` remains the authority, and no percentage is trustworthy until reconciled
+against it.
+
+Two surfaces, split on latency rather than taste:
+
+- `harness burn` (report, `weeks`, `calibrate`, `budget`, `reset-day`, `scan`,
+  `install`) — human-invoked, so the CLI's module graph is affordable.
+- `harness-burn-hud` (`line`, `session-start`, `stop`, `scan`) — a standalone binary
+  for the statusline repaint and the Stop hook. `harness --version` costs ~0.85s to
+  load against a ~0.11s repaint budget, so this binary imports nothing from
+  `@harness-engineering/*`; a test asserts that import graph, because the regression
+  would show up only as a terminal that feels slow.
+
+Every regression test from the Python suite came across, each still tied to a defect
+that actually shipped: the Monday-UTC week assumption that understated a 97% week by
+~81×, the write race that silently dropped 85% of the record store, transcript usage
+blocks inflating totals ~3.5×, and a 3-hour extrapolation firing CRITICAL. Parity was
+verified against 33,305 real records — every shared record byte-identical, and with
+`now` pinned the summaries differ only in float rendering.
+
+The port also fixed two hot-path defects of its own: the binary is emitted as `.mjs`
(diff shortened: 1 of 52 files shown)


chadjw:
The failing ci for windows looks legit. Small test failure.

chadjw (approved):

Issue #1146, opened 2026-08-07
check-docs/cleanup discovery: .mjs and dot-directories invisible, and zero files scanned reports 100%


ahhrealmonster:
## Summary

Two independent blind spots in the CLI's file discovery make `check-docs` and
`cleanup` report on a small, unrepresentative slice of a repo — and in the
degenerate case report a confident **100% green over zero files**.

Both were found by auditing why two gates in a consuming repo had never produced
a finding since install. That repo is an ESM, skills-and-docs overlay: 148 of its
186 source files are `.mjs`/`.cjs`, and its entire first-party surface lives
under dot-directories. `check-docs` scanned **11** of those 186 files — all of
them test fixtures — and reported `0.0%`.

Verified against `@harness-engineering/cli@10.2.0`.

## 1. Discovery is blind to `.mjs` / `.cjs` (`checkDocCoverage`)

```js
const sourceFiles = await findFiles("**/*.{ts,js,tsx,jsx}", sourceDir);
```

No `.mjs` / `.cjs`. Every ESM-first repo is invisible to docs coverage.

Note the entropy analyzer already gets this right — `DEFAULT_INCLUDE_PATTERNS`
lists `**/*.mjs` and `**/*.cjs`, and the JS parser claims `[".js", ".jsx",
".mjs", ".cjs"]`. Only `checkDocCoverage` disagrees.

## 2. Discovery never traverses dot-directories (`findFiles`)

Both variants call node-glob without `dot: true`:

```js
async function findFiles(pattern, cwd = process.cwd(), extraIgnore = []) {
  return glob(pattern, { cwd, absolute: true, ignore: [...] });   // dot: false
}
```

So anything under `.canary/`, `.harness/`, `.config/`, `.server/` … is never
seen. In the repo above that is *all* of the shipped code.

This one also silently corrupts **doc-drift**, which is the part that cost the
most time to diagnose. The exports index is built from the same discovery, so
symbols that plainly exist get reported as missing:

```console
$ harness cleanup --type drift
  - docs/wiki/Skills-Inventory.md: NOT_FOUND: Symbol "HttpClient" not found in codebase
  - docs/wiki/Skills-Inventory.md: NOT_FOUND: Symbol "AuthService" not found in codebase

$ grep -rn "export class HttpClient\|export class AuthService" .canary/skills/*/runtime/*.ts
  .canary/skills/<skill>/runtime/auth-service.ts:57:export class AuthService {
[...]

PR #1245, opened 2026-08-09
fix(graph): skip plain-git worktree containers during source walking

chadjw:
## Problem

`DEFAULT_SKIP_DIRS` skips agent sandbox dirs (`.claude`, `.cursor`, `.codex`, `.gemini`, …) because each can hold a full checkout of the repo. It missed the identical hazard created by `git worktree add .git-worktrees/<branch>` — reached without an agent.

`findFiles` sets `dot: true` deliberately (#1146) so first-party source under dot-directories stays visible, which makes this list **the only thing** keeping nested checkouts out. So every scanner re-walked each worktree as first-party source.

## Symptom

Three worktrees contributed **9,818** extra `.ts` files on top of the **3,302** real ones, surfacing as a phantom

```
[complexity] REGRESSION: 1327 > 333 (delta: 994)
```

with findings attributed to paths under `.git-worktrees/<branch>/packages/...`. This blocked commits that touched **no source at all** — it was hit while committing a docs-only roadmap change.

## Fix

Add `.git-worktrees` and `.worktrees` alongside the agent sandboxes, with a comment explaining why the list is load-bearing given `dot: true`.

## Verification

- 2 new regression tests mirroring the existing `.claude/worktrees` pair (assert membership + assert the walker does not descend). **13/13 pass** in `CodeIngestor-skip-dirs.test.ts`.
- With the fix built, `harness check-arch` exits **0** on a tree that previously failed; the blocked docs-only commit then went through.

## Assumptions made

- Included `.worktrees` as well as `.git-worktrees` — `.gitignore` already treated both as worktree containers, so leaving one out would have left the same bug reachable under the other name.
- Scoped to the shared skip-dirs set rather than per-scanner config, since the collectors that *did* behave correctly (`dep-depth`, `module-size`) do so via their own dot-dir guard; fixing the shared list repairs every consumer at once.

### DIFF
--- a/.antigravity-extension/config/mcp_config.json
+++ b/.antigravity-extension/config/mcp_config.json
@@ -2,7 +2,7 @@
   "mcpServers": {
     "harness": {
       "command": "npx",
-      "args": ["-y", "-p", "@harness-engineering/cli@11.0.0", "harness-mcp"]
+      "args": ["-y", "-p", "@harness-engineering/cli@11.1.0", "harness-mcp"]
     }
   }
 }
--- a/.changeset/ci-protected-main-writes.md
+++ b/.changeset/ci-protected-main-writes.md
@@ -1,0 +1,11 @@
+---
+'@harness-engineering/cli': patch
+---
+
+Fix CI's two paths for landing housekeeping commits on a protected `main`.
+
+`roadmap-auto-done.yml` declared `permissions: pull-requests: read`, but its branch-protection fallback calls `gh pr create` with the built-in `GITHUB_TOKEN`. That call failed with `Resource not accessible by integration (createPullRequest)` _after_ the branch had already been pushed, so every merge into a protected branch stranded a `chore/auto-done-pr*` branch and silently dropped its roadmap flip. The fallback was added without widening the permission, so it had never once succeeded — 58 stranded branches had accumulated, spanning a long run of merges.
+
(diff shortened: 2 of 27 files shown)


chadjw:
### Heads-up: the `build-and-test (ubuntu-latest, 22)` red here is **pre-existing on `main`**, not caused by this PR

The failing step is *"Baseline-gating regression test"*, and the actual assertion is `plugin-pin-sync`:

```
not ok 59 - every manifest pins @harness-engineering/cli to the CLI package version
  expected: '11.1.0'
  actual:   '11.0.0'
```

Verified against `origin/main` directly — the CLI is `11.1.0` there while all five plugin/extension manifests still pin `11.0.0`. So **any** branch cut from `main` inherits this failure; this PR merely happened to surface it.

Cause: `changesets/action@v1` is configured with `publish:` but no `version:` input, so it runs `changeset version` instead of `pnpm version` — and `pnpm version` is the only thing that invokes `scripts/sync-plugin-pin.mjs`.

Fixed in #1247 (stacked on this branch): passes `version: pnpm version` and re-pins the manifests. `node --test tests/scripts/*.test.mjs` goes 60 pass/2 fail → **62 pass/0 fail**.

Suggested merge order: **#1245 → #1247 → #1246**. The ruleset requires 1 approving review and defines no required status checks, so this PR is mergeable despite the inherited red; `main` goes green once #1247 lands.

PR #1292, opened 2026-08-10
feat(burn): attribute token spend to the subagent that spent it

### LINKED ISSUE #1270: Per-subagent token attribution in burn
Group burn's existing transcript scan by `attributionAgent` to produce per-subagent and per-fleet-lane token attribution. Claude Code writes one transcript per subagent to `~/.claude/projects/<project>/<sessionId>/subagents/agent-<id>.jsonl` (816 present locally), each carrying `agentId`, `attributionAgent`, `sessionId`, `sourceToolAssistantUUID`, `requestId`, `model` and a full `usage` block. `burn`'s `listTranscripts()` already recurses into those directories, so the data is ingested today and the identity discarded — this is a grouping key on a scan that already runs, plus the existing `requestId` dedup (transcripts repeat each usage block ~3x). Corrects a documented falsehood: `fleet-command/SKILL.md:319` states subagent tokens "are not observable, so a token governor would be a promise the skill cannot keep". Must assert the transcript shape and degrade to "unattributed" rather than 0 when the undocumented fields change, so a CLI update cannot silently report a fleet run as free. Unblocks per-lane cost measurement for Adaptive Model Routing (#1032). Source: paperclip budget-enforcement model (76.1k stars, MIT) — mechanism only, not the platform. Ideation: docs/ideation/external-source-adoption-tria-2026-08-09.md (score 9.00).
**Spec:** docs/changes/per-subagent-token-attribution-in-burn/proposal.md

chadjw:
## What

`burn` already reads every Claude Code transcript on this machine and already walks into
`~/.claude/projects/<project>/<sessionId>/subagents/` — subagent spend has always been in the
pooled totals. What was thrown away was *whose* spend it was. This adds the grouping key.

- `UsageRecord` gains `agent` and `agentId`. Labels are `main`, the subagent's
  `attributionAgent`, `unattributed`, or `pre-migration`.
- `Summary` gains additive `agents` and `attribution` blocks; `harness burn report` gains a
  **by agent** section.
- Subagent spend whose identity cannot be read is counted as **`unattributed` units** — never
  zero, never folded into `main`. When *none* of the week's subagent spend carries a readable
  label, the report headlines that attribution is degraded. A Claude Code update that renames
  the undocumented fields therefore cannot make a fleet run read as free.
- Corrects a documented falsehood in the shipped `fleet-command` skill body, which stated that
  subagent tokens "are not observable, so a token governor would be a promise the skill cannot
  keep". The design conclusion (the governor meters slots, not tokens) is kept — its premise is
  replaced with the true one: attribution is retrospective, so it cannot reserve before a
  dispatch happens.

## Shape assertion and degradation

A line is subagent spend if `isSidechain === true` **or** its file sits under a `subagents/`
path segment — two independent signals for one undocumented fact, so both must change at once
before classification degrades. Missing `attributionAgent` degrades to `unattributed` and the
units still count. The `unattributed` row is exempt from the report's top-N cut and unit floor,
because a small bucket that silently vanishes is the exact failure this exists to prevent.

## Store migration

`usage.tsv` widens 7 → 9 columns. `readRecords` previously discarded any row without exactly 7
fields, so a silent widening would have deleted the entire store. Instead: `files.tsv` gains a
`#version` header, a missing/older version forces one full rescan, legacy 7-column rows load as
`pre-migration`, and a narrow upgrade rule relabels any `pre-migration` row whose transcript is
still on disk. The reader accepts exactly 7 or 9-or-more fields, and both new columns are
sanitised on write.

Verified on the live HUD: 923/923 transcripts rescanned, store grew 40,047 → 40,552 records,
status never `UNDERCOUNT`, `pre_migration_units: 0` (every row healed), `degraded: false`, six
named agents visible.

## Assumptions made

Every default taken without a human in the loop, in one place:

1. **One autopilot phase, not four.** The spec's `## Implementation Order` lists four stages;
   they were planned and executed as a single phase with one plan artifact, because the stages
   are tightly coupled (the type change cannot land separately from its call sites).
2. **`agentId` is stored as well as `attributionAgent`.** The issue asks for per-subagent *and*
   per-fleet-lane attribution; `agentId` is the lane, and it is what makes a lane count
   computable. It is load-bearing, not speculative.
3. **No new `harness burn agents` subcommand and no statusline segment.** The report already
   has a "by model" section; the statusline runs against a ~0.11s budget.
4. **`pre-migration` is a distinct label from `unattributed`.** Labelling legacy rows
   `unattributed` would assert they were subagent spend (most are not) and would fire the
   degradation flag on history alone.
5. **Degradation is detected store-wide at summary time, scoped to the current week** — not
   from per-scan counters, which see only changed files and would read as an all-clear on a
   quiet scan.
6. **Rollback is stated, not mitigated.** A `burn` older than this change reading a 9-column
   store discards every row; the integrity gate then re-reads transcripts, so loss is bounded
   to rows whose transcripts are already pruned. Recorded in the spec and the changeset rather
   than solved with a sidecar file.
7. **Tab/CR/LF in the new columns is sanitised on write** rather than assumed absent.
8. **`fleet-command` is one real file, not four.** `agents/skills/{codex,cursor,gemini-cli}/fleet-command`
   are symlinks; the plan's original "copy over the other three" step was removed and replaced
   with an assertion that all three are still git mode `120000`.
9. **`agentsSection` was decomposed during execution** into `keptAgentLabels()` +
   `attributionCaution()` + `agentsSection()`. The plan's single function measured cyclomatic
   complexity 17 against the repo ceiling of 15, and `check-arch --update-baseline` refuses to
   record a genuine breach. Behaviour is unchanged and pinned by the plan's own tests.
10. **Internal roadmap prose is out of scope for the "no occurrence remains" criterion.**
    `docs/roadmap.md`, the roadmap shard and `docs/ideation/` also contain the phrase; they
    quote the defect and are not shipped skill bodies.

## Pipeline artifacts

- Spec: `docs/changes/per-subagent-token-attribution-in-burn/proposal.md`
- Plan: `docs/changes/per-subagent-token-attribution-in-burn/plans/2026-08-10-per-subagent-token-attribution-plan.md`
- A `harness-soundness-review --mode spec` pass produced 4 blocking findings (label collision,
  report elision, reader tolerance/rollback, the symlink error); all four were resolved in the
  spec and the plan was revised before execution.

Refs #1270

11. **A code-review pass ran after the executor and landed 4 further commits.** One is a
    genuine critical fix (below); the other three are triaged important/suggestion findings I
    accepted on the lane's own authority rather than a human's. They are reversible — the
    critical fix is the first of the four and is independent, so a reset back to it drops the
    other three cleanly if you disagree.
12. **Two review findings were deliberately NOT fixed in this PR.** (a) Success Criterion 10
    ("no occurrence of the claim remains") is verified by grep, not by a standing test — it is
    the one criterion without an executable guard. (b) `pad()` is 18 chars wide, so labels
    longer than that (e.g. `harness-task-executor`, 21) shift the units column; `modelsSection`
    sidesteps this by stripping a prefix. Both were judged not worth widening this diff.
13. **`packages/burn/tests/bin-startup.test.ts`'s timing test is load-sensitive and flaked
    locally.** It asserts a wall-clock ratio over 5 subprocess spawns. It is untouched by this
    PR (`git diff origin/main...HEAD` is empty for `src/bin/` and that test), it passed on all
    three CI OSes, and it passes locally with headroom once the machine is not saturated. The
    assertion that actually pins Success Criterion 9 is the separate deterministic bundle-grep
    test in the same file, which passes. Flagged for a future PR, not fixed here.

## Review findings

A `harness-code-review` pass found and fixed one **critical** defect: `toRecord` type-guarded
`attributionAgent` but took `agentId` verbatim, while `writeRecords` sanitises that column with
`String.replace`. A Claude Code release that changed the field's *type* — as plausible as a
rename, which the spec explicitly assumes — would have thrown out of `writeRecords`, aborted
the scan **before** the store was written, and repeated on every subsequent scan: a permanently
dead HUD. The lane id now degrades to empty, with an end-to-end regression test.

It also relocated two orphaned JSDoc blocks onto the functions they document, removed a new
cyclomatic-complexity violation in `readRecords` by extracting the row decode, bounded an
overstated forward-tolerance claim in a comment, and documented in the README that the
degradation flag detects *missing* labels but never *wrong* ones.

## Verification

`harness-verifier` independently re-ran every gate and reported **PASS, 11/11 success criteria,
0 gaps**. `harness validate` exits 1 with 410 issues both before and after this change —
byte-identical, none citing `packages/burn`, `packages/cli/src/commands/burn`, or the
fleet-command skill body.

Closes #1270


chadjw:
## Real-data verification (live HUD, final build)

Run against the actual transcript store on this machine, not a fixture — this is the roadmap-fleet batch that built the PR, attributing its own spend:

```
status: HOT
wtd:  6,513 requests · 151,644,461 units
scan: 928 transcripts, 928 rescanned (the one-time migration), 41,004 records

attribution:
  attributed_units      109,342,636
  main_units             42,301,825
  unattributed_units              0
  pre_migration_units             0
  lanes                          94
  degraded                    false

by agent:
  general-purpose               71,202,666   46 lanes   47.0%
  main                          42,301,825    0 lanes   27.9%
  harness-task-executor         20,366,064   12 lanes   13.4%
  harness-planner                7,019,729    8 lanes    4.6%
  harness-code-reviewer          5,171,164   11 lanes    3.4%
  harness-verifier               4,338,570   11 lanes    2.9%
  Explore                          459,525    3 lanes    0.3%
  harness-security-reviewer        417,437    2 lanes    0.3%
  harness-adversarial-reviewer     367,481    1 lane     0.2%
```

Four things worth checking against the spec:

1. **The partition is exact.** `attributed_units + main_units = 109,342,636 + 42,301,825 = 151,644,461`, which equals `wtd.units` to the unit — well inside the ±1-per-label tolerance Success Criterion 4 allows.
2. **The migration healed the whole store.** `pre_migration_units: 0` after one forced full rescan (928/928), and the store *grew* rather than shrank. `UNDERCOUNT` never fired.
3. **`degraded: false` with 94 lanes across 8 named agent types** — the shape assertion is reading the live transcripts correctly, not falling back.
4. **`main` reports 0 lanes**, as `AgentBlock.lanes` documents (main-thread records carry an empty `agentId`, which is excluded from the count).

This is the measurement the `fleet-command` prose correction now claims is possible.

PR #1485, opened 2026-08-24
feat(autopilot): general cross-skill lifecycle hooks (skillHooks) — supersedes review.additionalSkills (#1481)

### LINKED ISSUE #1481: harness-autopilot: REVIEW hardcodes one reviewer — let projects declare additional domain-specific review skills
## Summary

`harness-autopilot`'s `REVIEW` state hardcodes a single reviewer:

```
subagent_type: "harness-code-reviewer"
prompt: "Phase {N}: {name}. Session: {sessionSlug}. Follow harness-code-review..."
```

A project cannot declare additional, domain-specific review skills to run at
`REVIEW` or `FINAL_REVIEW`. The generic reviewer is good at generic findings —
but the risks that actually sink a given project are usually domain-shaped, and
there is no supported way to put a project's own detector in the loop.

## The mechanism already exists one layer up

Harness personas already model "a role composes several skills":

```yaml
# personas/security-reviewer.yaml
skills:
  - harness-security-review
  - harness-code-review
steps:
  - skill: harness-security-review
    when: on_pr
```

So the concept is present and expressed cleanly. `harness-autopilot` simply does
not consult it when dispatching `REVIEW` — the Persona Agents table maps each
state to one fixed `subagent_type`. The gap is the wiring, not the model.

## Motivating case

Building a detector whose entire purpose is catching false-green results
(closed issues whose fix never actually executed). The phases are heavily
TDD-driven, so the dominant project-specific risk is a **vacuous test** — one
that passes without proving anything. That project ships a skill which detects
exactly that.

The generic code reviewer is not tuned for it, and should not be: it is a
domain concern, which is precisely why the domain owner wants to inject it.

Wanted:

```jsonc
// harness.config.json
{
  "autopilot": {
    "review": {
      "additionalSkills": ["canary-cassandra"],
      "when": "per-phase"        // or "final" / "both"
    }
  }
}
```

## Why the workaround does not work

The obvious workaround is to record the extra reviewer in
`.harness/sessions/<slug>/autopilot-state.json`. That does not work, and the way
it fails is the clearest argument for the feature:

- The skill never reads such a key, so the entry is **inert**. It documents
  intent and changes nothing.
- The orchestrator must therefore dispatch the extra reviewer by hand at every
  phase boundary — which is exactly the "autopilot delegates, never
  reimplements" Iron Law being violated by the operator, because the skill left
  no other route.
- Because it is hand-dispatched, it is not covered by the skill's own gates.
  `REVIEW` cannot be skipped; a hand-run extra reviewer can be forgotten
  silently, which is the failure mode gates exist to prevent.
- State files are per-session and gitignored in most setups, so the policy does
  not travel with the repo. A second session, or another contributor, silently
  gets a weaker review.

That last point is the real cost: review policy that lives in a session file is
review policy that varies per operator without anyone noticing.

## Suggested shape

1. Read an `autopilot.review.additionalSkills` list from `harness.config.json`.
2. Run them at `REVIEW` after `harness-code-reviewer` (and optionally at
   `FINAL_REVIEW` via `when`), merging findings into
   `{sessionDir}/phase-{N}-review.json` under a per-skill key so provenance
   survives.
3. Treat a declared-but-unresolvable skill as a **failure, not a skip**. A
   review stage that silently drops a configured reviewer and reports green is
   the same false-green shape the review stage is there to prevent — it should
   surface as "cannot verify", consistent with how the rest of the harness
   handles abstention.
4. Keep `harness-code-reviewer` mandatory. This is additive; projects should not
   be able to swap out the baseline reviewer, only add to it.

Point 3 matters more than it looks: without it, a typo in a skill name
downgrades the review silently, and the config makes the project *feel* better
covered while covering it less.

## Related

- `harness-autopilot` SKILL.md — Persona Agents table, and the `REVIEW` /
  `FINAL_REVIEW` state definitions
- `personas/*.yaml` — the existing `skills:` composition model this would reuse

chadjw:
## Summary

Generalizes the (unreleased) one-off `review.additionalSkills` seam into a **general, cross-skill lifecycle-hook framework**: a project attaches additional **skills, commands, and prompts** at lifecycle points of *any* hook-supporting orchestrator skill via a new top-level `skillHooks` block in `harness.config.json`.

The narrow review case is now the flagship worked example: `skillHooks["harness-autopilot"]["after:REVIEW"]` + `["after:FINAL_REVIEW"]`. The `review.additionalSkills` field is **removed** (no back-compat baggage — it never shipped).

## Config shape

```jsonc
"skillHooks": {
  "harness-autopilot": {
    "before:EXECUTE": [
      "preflight-skill",                                   // bare string = skill shorthand
      { "type": "command", "run": "pnpm lint", "blocking": true },
      { "type": "prompt", "text": "Prefer existing helpers in packages/core/util." }
    ],
    "after:REVIEW":       [ { "type": "skill", "skill": "canary-cassandra", "blocking": true } ],
    "after:FINAL_REVIEW": [ "canary-cassandra" ],
    "on:failure":         [ { "type": "command", "run": "scripts/notify.sh" } ]
  },
  "harness-code-review": { "after:mechanical": ["extra-domain-check"] }
}
```

- **Outer key** = the hook-supporting skill's name (each skill declares its own event vocabulary; no universal enum).
- **Inner key** = event string, grammar `^(before|after|on):[A-Za-z0-9_-]+$` — phase boundaries (`before:/after:<phase>`), the whole invocation (`before:run`/`after:run`, so single-shot skills are hookable), and cross-cutting lifecycle events (`on:failure`, `on:park`, …).
- **Value** = ordered array of entries: a bare skill-name string or a discriminated `{type: "skill" | "prompt" | "command"}` object, each with an optional `enabled` toggle.

## Three hook kinds

- **`skill`** — dispatch as an additional subagent (LLM path); brief carries the hook context.
- **`prompt`** — mechanically append `text` to the phase's persona prompt (declarative, never blocks).
- **`command`** — mechanically run a shell command via the command-runner (deterministic, no LLM), receiving the context as `HARNESS_*` env vars + JSON on stdin.

## Invariants preserved (carried forward, not weakened)

- **Hard halt = false-green protection**: an unresolvable/undispatchable skill, or a command that **cannot be spawned**, is a hard halt (recorded in `.harness/failures.md`, not routed through fix/override/stop). A command that **ran and exited non-zero** is a normal blocking finding (not a halt).
- Default-blocking policy: review/verify events default `blocking:true`, else `false`; a per-entry `blocking` overrides. `prompt` never blocks.

## Shared code (real seam, not just prose)

- `@harness-engineering/core` → `packages/core/src/hooks/`: `resolveSkillHooks(config, skillName, event) → NormalizedHook[]`, `defaultBlocking(event)`, and the hook input-context helpers (`buildHookEnv` / `buildHookStdinPayload` / `buildHookBriefLines`). Unit-tested directly for both consumers, all three kinds, the `enabled` toggle, grammar, and empty/absent config.

## Two proven consumers

1. **harness-autopilot** (reference): documents the full event vocabulary + one generic dispatch pattern; proven across `after:REVIEW` + `after:FINAL_REVIEW` (migrated), `before:EXECUTE` (non-review generality), and `on:failure` with `HARNESS_FAILURE_REASON`, over all three kinds.
2. **harness-code-review** (second consumer, proves it's not autopilot-locked): declares `after:mechanical`; resolver-level proof.

## Follow-ups

- Wire the remaining hook-supporting skills (brainstorming, fleets, …) — the extension contract is documented.
- RESERVED v2 (documented, not implemented): per-iteration granularity (`after:EXECUTE:task`, `after:dispatch:item`), a `"*"` wildcard outer key, and `{{token}}` prompt templating. (Conditional scoping works today via a command self-gating on `$HARNESS_CHANGED_FILES`.)

Closes #1481


chadjw:
## Code-review findings addressed — pushed `1b6ee7221`

Ran the `harness-code-review` pipeline on this PR (1 important + 4 suggestions, 0 critical). All addressed on-branch:

| # | Sev | Finding | Fix |
|---|-----|---------|-----|
| 1 | **important** | `harness-autopilot` SKILL.md advertised `before/after` for all 6 phase STATEs, but only **4** hook points are wired (`before:EXECUTE`, `after:REVIEW`, `after:FINAL_REVIEW`, `on:failure`). A hook at an unwired event was a **silent no-op** — the exact false-green the framework exists to prevent. | Narrowed the vocabulary claim to the wired set; marked the rest **"not yet wired"** (documented follow-up). Regenerated the gemini/antigravity autopilot command artifacts. Added a **mechanical backstop test** asserting the documented wired set equals the concrete `resolveSkillHooks(...)` call sites, so it can't silently overpromise again (this also gives the otherwise prose-only hard-halt invariant a guard). |
| 2 | suggestion | `defaultBlocking` used a substring match (`/review\|verify/i`), so `after:preview` → `blocking:true` (also `reverify`/`unverified`). | Tokenize the state on `[_-\s]` and match **whole tokens**; `FINAL_REVIEW` still blocks, `preview` no longer does. Added collision tests. |
| 3 | suggestion | Core `SkillHookEntry` accepted a typeless `{skill}` shorthand (with a passing test), but the CLI zod schema (`discriminatedUnion('type')`) **rejects** it — two sources of truth disagreed, and it was unreachable via validated config. | Dropped the typeless branch + its test; converted the test to an explicit `{type:'skill'}` default-blocking case. |
| 4 | suggestion | `proposal.md` + `plan.md` still described the removed `review.additionalSkills` design (their acceptance criteria would no longer parse). | Added a **"Superseded (PR #1485)"** banner to both pointing at `configuration.md` + the changeset. |
| 5 | suggestion | Duplicated `violationId` in the arch allowance file. | Deduped. |

**Verification (Node 22):** core hooks tests 26/26, CLI schema+integration 28/28, `typecheck` core+cli clean, `prettier --check` clean, `generate:plugin:check` OK, pre-push gauntlet (reference-docs + tool-catalog) fresh.

**Observation (no fix, by design):** the hard-halt false-green protection is enforced in SKILL.md prose executed at runtime; the integration test greps the markdown for it rather than executing a halt. Inherent to prose-driven skills — flagged so it's on the record; the new call-site backstop (Finding 1) is the closest mechanical guard.

PR #1703, opened 2026-08-28
feat(pricing): add cheapest-model-by-cost helper [A/B: comprehension ON]

chadjw:
## A/B test — Run A (comprehension ON)

Adds `cheapestModelByCost(models, dataset)` to `packages/core/src/pricing` — returns the model id with the lowest combined input+output cost per 1k tokens from a candidate list. Absent models are skipped; ties favor the earlier model; returns `null` when none are priced.

- New: `packages/core/src/pricing/select.ts` + unit test (6 cases, all green)
- Exported from the pricing barrel
- Matches module conventions (per-million normalization, exact-match lookup like `getModelPrice`)

**This is one arm of a two-arm comprehension A/B measurement — not intended to merge.** Context for the build was gathered via the compiled comprehension unit (`get_comprehension` serve path) for `packages/core/src/pricing`. See the paired baseline PR and the comprehension audit for the token comparison.

Comprehension side effect: editing the module invalidated its committed unit; the pre-commit hook recompiled it **static-only** (no semantic provider in-session), which is why `_module.md` shrinks in the diff.

### DIFF
--- a/.changeset/pricing-cheapest-model.md
+++ b/.changeset/pricing-cheapest-model.md
@@ -1,0 +1,5 @@
+---
+'@harness-engineering/core': minor
+---
+
+pricing: add `cheapestModelByCost` helper that returns the model id with the lowest combined input+output cost per 1k tokens from a candidate list.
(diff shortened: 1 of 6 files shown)


chadjw:
Closing — throwaway comprehension A/B measurement PR (never intended to merge). The token comparison is captured; the degradation it surfaced is fixed forward by ADR 0109 (#1705) and the follow-up slices.

Issue #1811, opened 2026-09-01
fix: assignment-history table silently drops records whose feature name contains a pipe (round-trip data loss)


chadjw:
## Summary

The roadmap **Assignment History** table loses records on a `parse → serialize → parse` round-trip whenever a feature name (or assignee) contains a pipe (`|`). The record is **silently dropped** — no error, no warning.

Discovered by a proactive `bug-fleet` sweep. Reproduction is deterministic and confirmed **red by assertion** at base `f3323ea1156353e2ce3e775dc19ae84d6702b697`.

## Root cause

`serializeAssignmentHistory` emits each record as a markdown pipe-table row (`packages/core/src/roadmap/serialize.ts`):

```
| ${record.feature} | ${record.assignee} | ${record.action} | ${record.date} |
```

`parseAssignmentHistory` reads it back (`packages/core/src/roadmap/parse.ts`) by splitting the row on `|` and dropping empty cells:

```ts
const cells = trimmed.split('|').map((c) => c.trim()).filter((c) => c.length > 0);
if (cells.length < 4) continue;
const action = cells[2] as AssignmentRecord['action'];
if (!['assigned', 'completed', 'unassigned'].includes(action)) continue;
```

No cell is escaped. A feature name such as `Auth | Login flow` serializes to a row with an extra `|`, so on reparse the cells shift: `action` lands on `cells[2] = 'flow'`, which is not a valid action, and the entire record is skipped (dropped).

Feature names are free text (H3 headings / the MCP `manage_roadmap` write path), so a name containing `|` is reachable in normal use.

This is the **same class** of line-oriented round-trip data loss already fixed for the comma-in-list field (#1757) and the newline-in-summary field (#1756). Those fixes deliberately scoped themselves to the single-line `- **Field:**` bullets and did not cover the Assignment History pipe table.

## Reproduction

A red-at-base regression test is on branch [`repro/roadmap-assignment-history-pipe`](https://github.com/Intense-Visions/harness-engineering/tree/repro/roadmap-assignment-history-pipe) at `packages/core/tests/roadmap/repro-assignment-history-pipe.test.ts`:

```ts
it('preserves an assignment record whose feature name contains a pipe', () => {
  const roadmap = structuredClone(VALID_ROADMAP);
  roadmap.assignmentHistory = [
    { feature: 'Auth | Login flow', assignee: 'alice', action: 'assigned', date: '2026-03-21' },
  ];
[...]

Issue #543, opened 2026-06-04
Promote harness:strategy to gateway position in init


chadjw:
`initialize-harness-project/SKILL.md` currently asks the strategy question as Phase 3, step 5c — buried after scaffolding and configuration. The article's gear item #1 is "specs operated FROM," and strategy is the foundational spec. Move the strategy prompt to the FIRST question init asks, not the fifth. Adopters who skip it end up with no strategic anchor, which means brainstorming/ideate/roadmap-pilot start cold. Source: Pass 3 #13.
**Spec:** docs/changes/strategy-init-gateway/proposal.md

chadjw:
## Harness Analysis: promote-harness-stra-2b2640f5

**Risk:** low (80% confidence)
**Route:** local
**Analyzed:** 2026-06-05T01:33:18.744Z

- Structural complexity: 0.02 (13 files, 0 services, 0 modules affected)
- Semantic complexity: 0.40 (2 unknowns, 2 ambiguities, 1 risk signals)
- Historical complexity: 0.00 (from past execution outcomes)
- Overall: 0.15 → risk level "low", recommended route "local"

<details>
<summary>Full Analysis Data</summary>

```json
{
  "_harness_analysis": true,
  "_version": 1,
  "issueId": "promote-harness-stra-2b2640f5",
  "identifier": "promote-harness-stra-2b2640f5",
  "spec": {
    "id": "promote-harness-stra-2b2640f5",
    "title": "Promote harness:strategy to gateway position in init",
    "intent": "Reorder the initialization workflow to prioritize collecting the core strategy question at the beginning.",
    "summary": "Move the strategic positioning/question from its current location (Phase 3, step 5c) within the `initialize-harness-project` guide to be the very first question presented to users during project initialization. This ensures that new adopters establish a foundational strategic anchor before proceeding with other setup steps.",
    "affectedSystems": [
      {
        "name": "Initialization Workflow",
        "graphNodeId": "module:packages/orchestrator/tests/workflow",
        "confidence": 0.7,
        "transitiveDeps": [
          "file:packages/orchestrator/tests/workflow/config.test.ts",
          "file:packages/orchestrator/tests/workflow/loader.test.ts",
          "file:packages/orchestrator/tests/workflow/routing-config-schema.test.ts",
          "file:packages/orchestrator/tests/workflow/routing-cross-field.test.ts",
          "file:packages/orchestrator/tests/workflow/routing-isolation-schema.test.ts",
[...]

PR #1854, opened 2026-09-05
fix(core)!: reject path-traversing External-IDs and re-validate at every authenticated GitHub sink

### LINKED ISSUE #1843: External-ID regex in core admits path traversal into an authenticated api.github.com sink (second instance of #1842)
Filed by `fleet-command` (2026-09-05 run, wave 2) under explicit human authorization. Reproduced by the `security-fleet` lane; **parked rather than fixed** because the authorized fix scope for the sibling instance was deliberately kept minimal to one call site.

## Second live instance of the class fixed in #1842

`packages/core/src/roadmap/external-id.ts:12`:

```
const EXTERNAL_ID_RE = /^github:([^/]+)\/([^#]+)#(\d+)$/;
```

The `repo` capture group `[^#]+` admits both `/` and `?`. `parsed.repo` is then interpolated **unencoded** into a token-bearing `api.github.com` sink — `packages/core/src/roadmap/adapters/github-issues.ts:475`, alongside `Authorization: Bearer ${this.token}` at `:201` — across roughly 9 sibling call sites.

## Reproduction

Run against the verbatim regex and template; 3 of 4 probes escape the intended path:

| `External-ID` value | resolves to |
| --- | --- |
| `github:x/../../../user/emails?#1` | `POST https://api.github.com/user/emails` |
| `github:a/../../../applications/CLIENTID/token?#1` | `POST https://api.github.com/applications/CLIENTID/token` |

The trailing `?` truncates the intended path suffix into a query string, and `/../` collapses under WHATWG URL normalization.

Source is the `External-ID` field of any of the 283 `docs/roadmap.d/*.md` shards, which are fleet-written and PR-contributable.

## Why this is worse than #1842, not a duplicate of it

#1842 hardened **one** call site in `packages/dashboard/src/server/routes/actions.ts`. This file's own docstring calls itself *"The one regex that defines the External-ID format"*, and ADR `0051-slug-identity-external-id-sync-key.md` names `parseExternalId` / `buildExternalId` **"the single format authority."**

So the authoritative, permissive regex remains in force everywhere else, and the two are now genuinely out of sync — the hardened copy is the exception rather than the rule. Independently checked by the `adr-fleet` lane, which classified this as a **defect against an already-decided contract** (ADR 0051), not a new decision.

## Scope, stated honestly

The host is a fixed literal and normalization cannot change it, so this is **not** token exfiltration to an attacker-controlled host. The primitive is **method + path control against `api.github.com` using the operator's ambient token**.

## Suggested remedy

Mirror #1842: constrain the regex to `^github:([^/#?\s]+)\/([^/#?\s]+)#(\d+)$`, add a dot-segment guard (`encodeURIComponent` does **not** encode `.`, so `github:../..#1` still escapes under the constrained regex alone — this was found the hard way in #1842), and percent-encode each path segment at the sink.

Refs #1842. Related: ADR 0051.

chadjw:
Closes #1843

Second live instance of the class fixed for the dashboard in #1842 — this one in `packages/core`, the format authority ADR 0051 names. Branched from `origin/main` `5cd661d74`, **not** from #1842, and it does not touch #1842's dashboard change.

## The defect

`packages/core/src/roadmap/external-id.ts:12` defined the canonical format as:

```
/^github:([^/]+)\/([^#]+)#(\d+)$/
```

Those captures are "everything up to the next delimiter", not GitHub's owner/repo name grammar. The owner capture admitted `.`, `..` and `?`; the repo capture additionally admitted `/`. Ten adapter call sites then spliced them, **unencoded**, into `${apiBase}/repos/${owner}/${repo}/issues/${n}/...` on a request carrying `Authorization: Bearer <token>`.

Reproduced at `5cd661d74` against the verbatim regex and the verbatim URL template (Node 22) — 3 of 4 probes escape `/repos/`:

| External-ID | resolves to |
| --- | --- |
| `github:x/../../../user/emails?#1` | `POST https://api.github.com/user/emails` |
| `github:a/../../../applications/CLIENTID/token?#1` | `POST https://api.github.com/applications/CLIENTID/token` |
| `github:o/../../../user/repos?#1` | `POST https://api.github.com/user/repos` |
| `github:Intense-Visions/harness-engineering#1843` | *(contained — resolves correctly)* |

The dot segments collapse under WHATWG URL normalization and the trailing `?` truncates the intended `/issues/<n>/assignees` suffix into a query string. The source is the `- **External-ID:**` field of any `docs/roadmap.d/*.md` shard, which is fleet-written and PR-contributable.

Scope, stated honestly: `apiBase` is a fixed literal and normalization cannot change the host, so this is **not** token exfiltration to an attacker-controlled host. The primitive is **method + path control against `api.github.com` using the operator's ambient token**.

## The fix — both layers

The fork was answered **(c) BOTH**, and both halves land.

**Layer 1 — tighten the authority.** `EXTERNAL_ID_RE` is now held to GitHub's own name grammar (owner `[A-Za-z0-9][A-Za-z0-9-]{0,38}`, repo `[A-Za-z0-9._-]{1,100}`), plus an explicit dot-segment guard in `parseExternalId`. The guard is separate on purpose: `encodeURIComponent` leaves `.` untouched, so `..` survives both the character class and the encoding — the lesson #1842 learned the hard way.

**Layer 2 — re-assert at the sink.** New `githubRepoPath(owner, repo)` export re-validates and percent-encodes. It **never consults the regex**, so it still holds if the pattern is loosened again. It is called at all ten call sites immediately before the URL is built:

- `roadmap/adapters/github-issues.ts` — `updateTicket`, `fetchTicketState`, `assignTicket`, `addComment`, `fetchComments`/`fetchCommentPage`
- `roadmap/tracker/adapters/github-issues.ts` — `fetchById`, `update`, `fetchRawLabels`, `appendHistory`, `fetchHistory`

Only two of those ten (`fetchById`, `update`) had any incidental protection before, via an `parsed.owner !== this.owner` equality check. The other eight had none. All ten already sat inside `try { ... } catch { return Err(...) }`, so a rejection degrades to exactly the `Err` the existing `if (!parsed)` guard produced — no caller contract changes.

## Breaking change — explicitly authorized

An `External-ID` that GitHub itself could never have issued is now **rejected** rather than parsed. The human considered a compatibility/migration check at the CONFIRM gate and **deliberately dropped it**; no compat shim was reinstated.

**Shard-rejection sweep (the requested finding).** I transcribed the *shipped* validation — `EXTERNAL_ID_RE` + `isDotSegment` + `githubRepoPath`, all three — and ran it over every `- **External-ID:**` value in `docs/roadmap.d/*.md`:

- 282 shards scanned (283 files minus `_meta.md`)
- 280 carry a real External-ID; 2 carry an em-dash placeholder; 0 have no field
- **0 shards rejected.** No shard is named because none failed.

Widened to a repo-wide scan of **538** distinct `github:...#N` literals across `packages/` and `docs/`: the only rejections are the 7 attack probes this PR's own test introduces, plus `docs/changes/strategy-init-gateway/proposal.md:69` `` `github:...#543` `` — a prose ellipsis in a proposal, not a live External-ID.

No `docs/roadmap.d/*.md` file is edited by this PR (`git status -- docs/roadmap.d` → 0), leaving that surface clear for #1811.

## Reproducing test

`packages/core/tests/roadmap/external-id-path-traversal.test.ts` (223 lines, 80 cases). It exercises the **traversal** — it drives the real adapters with an injected recording `fetchFn` and asserts on the URL actually handed to `fetch`, resolved through `new URL()`, not on the shape of the regex. `/repos/x/../../../user/emails?…` passes a naive prefix check and still resolves to `/user/emails`; the assertion is that `new URL(url).pathname` still starts with `/repos/` and contains `/issues/`. Both defence layers are covered independently, so a regression in either one fails on its own.

```
cd packages/core && npx vitest run tests/roadmap/external-id-path-traversal.test.ts
```

| | result |
| --- | --- |
| before the fix | **69 failed \| 11 passed (80)** |
| after the fix | **80 passed (80)** |
| revert-and-fail (`git stash push -- packages/core/src`, test kept) | **69 failed \| 11 passed (80)**, then 80 passed on restore |

## Verification (Node v22.23.2, the `.nvmrc` pin)

- `packages/core` full suite — 466 passed \| 1 skipped (467 files), **5124 passed** \| 1 skipped
- `packages/dashboard` — 134 files, **950 passed**
- `packages/orchestrator` — 260 passed \| 1 skipped (261 files), **2883 passed** \| 1 skipped. One `EADDRINUSE 127.0.0.1:54643` ephemeral-port flake outside the diff on the first run; clean on rerun.
- `pnpm turbo typecheck lint` — 34/34 successful
- `pnpm turbo build` — 13/13 successful
- `scripts/generate-core-barrel.mjs --check` — up to date
- Pushed through the full pre-push gauntlet (changeset, whole-tree `format:check`, coverage ratchet, reference-docs freshness, tool-catalog) with **no** `--no-verify`.

## Assumptions made

1. **F1 was answered (c) BOTH** by the human at the fleet CONFIRM gate — tighten the regex *and* add a defence-in-depth assertion at the `api.github.com` sink. Neither half was treated as optional or as a substitute for the other.
2. **Breaking was explicitly authorized.** A compatibility/migration check was considered and deliberately dropped, so the stricter regex ships without one. The shard sweep above is reported as a **finding** (0 rejections), not enforced as a gate.
3. **Scope boundary honoured:** `docs/roadmap.d/*.md` was read for the sweep but never edited — #1811 is being built on top of this branch and owns `_meta.md`.
4. **#1842 was not branched from and its dashboard change is not duplicated.** Base is `origin/main` `5cd661d74`.
5. `parseRepoParts`-derived `this.owner`/`this.repo` come from operator-supplied `harness.config.json`, a different trust tier from PR-contributable shard content, so they were left unchanged.
6. `packages/orchestrator/src/core/pr-detector.ts:43` re-implements its own permissive parse, but it feeds no authenticated path sink. Noted as deferrable follow-up, not fixed here.
7. A `bug` route leaves **no** `plans/` directory — expected, not a defect. The committed pipeline artifacts are `docs/changes/external-id-path-traversal-1843/provenance.json` and `debug-session.md` (the live session under `.harness/debug/` is gitignored by `.harness/.gitignore:4`, so a copy is committed for verifiability).

Refs #1842. Related: ADR `0051-slug-identity-external-id-sync-key.md`.

### DIFF
--- a/.changeset/plain-pandas-repeat.md
+++ b/.changeset/plain-pandas-repeat.md
@@ -1,0 +1,24 @@
+---
+'@harness-engineering/core': patch
+---
+
+fix(core)!: constrain the canonical External-ID regex and re-validate at every authenticated GitHub API sink
+
+`parseExternalId` — the format authority named by ADR 0051 — matched
+`^github:([^/]+)\/([^#]+)#(\d+)$`. The owner capture admitted `.`, `..` and `?`; the repo
+capture additionally admitted `/`. Ten adapter call sites spliced those captures,
+unencoded, into `${apiBase}/repos/${owner}/${repo}/issues/${n}/...` on requests carrying
+`Authorization: Bearer <token>`, so a crafted `External-ID` in a PR-contributable roadmap
+shard could choose the path of a credentialed request — `github:x/../../../user/emails?#1`
+resolved to `POST https://api.github.com/user/emails`.
+
+Owner and repo are now held to GitHub's own name grammar, bare dot segments are rejected
+separately (`encodeURIComponent` does not encode `.`), and a new `githubRepoPath` export
+re-asserts and percent-encodes at each sink independently of the regex, so a future
+loosening of the pattern cannot silently re-open the traversal.
+
+**Breaking:** an `External-ID` that GitHub itself could never have issued is now rejected
+rather than parsed. No committed roadmap shard is affected — all 280 shards carrying a real
+`External-ID` still parse.
+
+Second instance of the class fixed for the dashboard in #1842.
--- a/.changeset/tame-hounds-shave.md
+++ b/.changeset/tame-hounds-shave.md
@@ -1,0 +1,29 @@
+---
+'@harness-engineering/core': minor
+---
+
+fix(core): stop losing assignment-history records whose feature name contains a pipe
+
+The roadmap's `## Assignment History` section was emitted as a markdown pipe table and
+read back by recovering its four values POSITIONALLY from an unescaped `split('|')`.
+Nothing escaped the separator, and a feature name is free text (an H3 heading, or the
+MCP `manage_roadmap` write path). So a name such as `Auth | Login flow` serialized to a
+row with five cells, `action` landed on the wrong cell, failed its membership check, and
+the WHOLE record was dropped on the next read — silently, with no error and no warning
+(#1811).
+
+The column separator is now gone rather than escaped. Each record is written as a block
+of four `- **Key:** value` bullets — the same line grammar every feature row already
+uses — so `|` has no special meaning and cannot shift a value onto the wrong field.
+Newlines are handled by the existing summary-field codec; backticks, em-dashes and
+table-separator lookalikes are inert. `serialize → parse` is now an identity for every
+field value, including leading and trailing whitespace, which the table used to trim.
+
+Reading is backward-compatible: the legacy pipe table is still parsed, with its original
+tolerances, so a `roadmap.md` or `_meta.md` written before this change keeps its history
+instead of losing it on first read. Only the writer moved. The first re-serialization of
+a document migrates it. A legacy row whose value contained a `|` was already destroyed
+when it was written and cannot be recovered.
+
+This repo's own `docs/roadmap.d/_meta.md` and the regenerated `docs/roadmap.md` are
+migrated in this change, with all 18 existing records preserved.
--- a/.harness/comprehension/packages/core/src/roadmap/_module.md
+++ b/.harness/comprehension/packages/core/src/roadmap/_module.md
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1
 module: 'packages/core/src/roadmap'
-sourceHash: '7c66b80fdddef0e3cba73dfaa49e1220060d2677e74a9e2f15eb489e9a129fcf'
+sourceHash: '9b50ca192f87e3de9ed816ef0112fe3a5ef9ee67425faaf32f6790fdd458901a'
 compiler: { static: '1.0.0', semantic: '1.0.0' }
 model: null
 semantic: absent
@@ -108,6 +108,7 @@ export detectRoadmapStorageMode
 export fullSync
 export getRoadmapMode
 export getTrackerKindRegistration
+export githubRepoPath
 export groomRoadmap
 export isClaimableBy
 export isMachineAssignee
--- a/.harness/comprehension/packages/core/src/roadmap/_module.md
+++ b/.harness/comprehension/packages/core/src/roadmap/_module.md
@@ -1,13 +1,14 @@
 ---
 schemaVersion: 1
 module: 'packages/core/src/roadmap'
-sourceHash: '9b50ca192f87e3de9ed816ef0112fe3a5ef9ee67425faaf32f6790fdd458901a'
+sourceHash: '4c61914f56c6b8ae13338b9417230e83301ce88caa3f3cb20d036d473d07c839'
 compiler: { static: '1.0.0', semantic: '1.0.0' }
 model: null
 semantic: absent
 members:
   [
     'assignee-lifecycle.ts',
+    'assignment-history.ts',
     'derive-repo.ts',
     'external-id.ts',
     'heading.ts',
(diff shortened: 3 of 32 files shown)


chadjw:
CI note — the first `build-and-test (macos-latest, 22)` attempt failed on `packages/dashboard/tests/client/pages/Impact.test.tsx > fetches anomalies on mount and renders articulation points sorted by dependentCount desc` (`TestingLibraryElementError: Unable to find role="heading" and name /Articulation Points/i`).

Proven outside the diff before rerunning, not rubber-stamped:
- `gh pr view 1854 --json files` lists **zero** `packages/dashboard` paths — this PR touches only `packages/core/src/roadmap/*`, one core test, the changeset, the provenance artifacts, and four generated comprehension shards.
- The failing assertion is a React Testing Library async-render timing check with no dependency on `parseExternalId` / `githubRepoPath`.
- That file was re-run **3/3 green locally** at this branch head (Node v22.23.2), and the full `packages/dashboard` suite is 950/950 green locally.

`gh run rerun 33985536965 --failed` → macOS now passes (10m45s). All three OSes green: ubuntu pass, windows pass, macos pass.

PR #1861, opened 2026-09-06
chore(config): enable github and exa MCP integrations; record security posture

chadjw:
Commits the working-tree changes left over after the 2026-09-05 `fleet-command` run, once `main` was synced to `180ebbb9c`.

## What's in it

**MCP server set** — `.mcp.json` and `.gemini/settings.json`, kept in sync:
- remove `sequential-thinking`
- add `playwright`, `github`, `exa`
- credentials are `${VAR}` references only — no literal secrets

**`harness.config.json`** — declare `integrations.enabled` = `["github", "exa"]` to match.

**`.harness/security/timeline.json`** — a 104-line security-posture snapshot auto-appended by `harness check-security` during the run's `security-fleet` lane. Records `commitHash 5cd661d74`, 6 findings (0 error / 5 warning / 1 info). The file had carried no entry since 2026-07-13.

All four files had been reformatted by a non-prettier writer (arrays expanded to multiline). `prettier --write` normalized that back, so the diff shows only semantic change.

## Deliberately excluded — and worth a look

The post-checkout hook regenerated `docs/roadmap.md` and the result **deletes the entire 92-line `## Assignment History` section**. That regeneration is not in this PR.

Cause: #1859 converted assignment history in `docs/roadmap.d/_meta.md` from a pipe table to a list. The globally-installed `harness` is the published **12.3.0**, which predates that change and still parses the old table format. It reads the new list, gets **0 records**, and `serializeRoadmap` (`packages/core/src/roadmap/serialize.ts:58`, "omit if empty") drops the section.

- Source data is **intact** — 18/18 records still in `docs/roadmap.d/_meta.md`
- Only the *regenerated aggregate* loses them
- Any user on the released CLI who runs `harness roadmap regen` before 12.4.0 ships will silently delete that section

Filed separately rather than papered over here.

### DIFF
--- a/.gemini/settings.json
+++ b/.gemini/settings.json
@@ -8,13 +8,23 @@
       "command": "npx",
       "args": ["-y", "@upstash/context7-mcp"]
     },
-    "sequential-thinking": {
-      "command": "npx",
-      "args": ["-y", "@modelcontextprotocol/server-sequential-thinking"]
-    },
     "playwright": {
       "command": "npx",
       "args": ["-y", "@playwright/mcp"]
+    },
+    "github": {
+      "command": "npx",
+      "args": ["-y", "@modelcontextprotocol/server-github"],
+      "env": {
+        "GITHUB_PERSONAL_ACCESS_TOKEN": "${GITHUB_PERSONAL_ACCESS_TOKEN}"
+      }
+    },
+    "exa": {
+      "command": "npx",
+      "args": ["-y", "exa-mcp-server"],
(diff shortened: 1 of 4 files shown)


chadjw:
The excluded `docs/roadmap.md` regeneration is filed as #1862 — `harness roadmap regen` on the published 12.3.0 CLI silently deletes the 92-line Assignment History section after #1859's format migration. Source data in `_meta.md` is intact.

Issue #1882, opened 2026-09-06
test-craft: --no-source-pair is inert — same Commander negated-flag inversion as the design-pipeline sibling


chadjw:
## Summary

`harness test-craft` declares `--no-source-pair` to skip source-pairing resolution. **It is inert.** Passing it changes nothing; source pairing still runs.

## Evidence (at `origin/main` `c1ca02ba2`)

Declared — `packages/cli/src/commands/test-craft.ts:40`:

```ts
.option('--no-source-pair', 'Skip source-pairing resolution')
```

Read — `packages/cli/src/commands/test-craft.ts:86` (inside `buildInput`):

```ts
if (opts.noSourcePair === true) input.sourcePair = false;
```

## Root cause

Identical to the sibling issue in `design-pipeline.ts`. For `.option('--no-source-pair', ...)`, Commander populates **`opts.sourcePair`** (default `true`, set to `false` when the flag is passed). It never creates `opts.noSourcePair`.

`opts.noSourcePair` is therefore `undefined` on every path, `input.sourcePair` is never set to `false`, and source pairing runs unconditionally.

Note the surrounding lines in the same `buildInput` function (`opts.files`, `opts.frameworks`, `opts.maxFiles`, `opts.maxTestsPerFile`, `opts.emit`) are all correct — they are plain non-negated options. Only the single `--no-x` flag is wrong.

## Impact

Lower blast radius than the `design-pipeline` sibling (source pairing is a best-effort enrichment, not a phase gate), but the same silent-no-op shape: a documented flag that reports nothing and does nothing. It also means the "skip source pairing" escape hatch is unavailable on repos where pairing is slow or noisy — the documented remedy simply does not work.

Same class as #1819 and #1345.

## Suggested fix

```ts
if (opts.sourcePair === false) input.sourcePair = false;
```

Also correct `TestCraftCliOptions`, which declares `noSourcePair?: boolean` — a shape Commander never produces. As with the sibling, the interface is what let the wrong read typecheck cleanly, so the type declaration is part of the defect rather than incidental to it.

## Regression test

An option-parsing-boundary test asserting `['--no-source-pair']` produces `input.sourcePair === false`, and that its absence leaves pairing enabled.

Note `packages/cli/tests/**` is region-locked by an in-flight PR at time of filing; a colocated `packages/cli/src/commands/test-craft.*.test.ts` is an established alternative in this repo.

## Provenance

[...]

PR #1953, opened 2026-09-06
fix(cli): make design-pipeline --no-freshen and --no-fill actually skip their phases

### LINKED ISSUE #1881: design-pipeline: --no-freshen and --no-fill are inert — Commander populates opts.freshen/opts.fill, the code reads opts.noFreshen/opts.noFill
## Summary

`harness design-pipeline` declares two documented skip-flags, `--no-freshen` and `--no-fill`. **Both are permanently inert.** Passing them changes nothing; the FRESHEN and FILL phases run anyway.

## Evidence (at `origin/main` `c1ca02ba2`)

Declared — `packages/cli/src/commands/design-pipeline.ts:45-46`:

```ts
.option('--no-freshen', 'Skip the FRESHEN phase')
.option('--no-fill', 'Skip the FILL phase (input bootstrap + craft polish)')
```

Read — `packages/cli/src/commands/design-pipeline.ts:62-63`:

```ts
if (opts.noFreshen === true) input.noFreshen = true;
if (opts.noFill === true) input.noFill = true;
```

## Root cause

This is Commander's negated-flag convention, applied backwards.

For `.option('--no-freshen', ...)`, Commander does **not** create an `opts.noFreshen` key. It creates `opts.freshen`, defaults it to `true`, and sets it to **`false`** when the flag is passed. The same for `--no-fill` → `opts.fill`.

So `opts.noFreshen` and `opts.noFill` are `undefined` on every code path — with the flag, without the flag, always. `undefined === true` is never satisfied, `input.noFreshen` / `input.noFill` are never set, and the pipeline runs both phases unconditionally.

The `--fix` and `--ci` reads immediately above and below these two lines (`opts.fix === true`, `opts.ci === true`) are correct, because those are plain non-negated flags. Only the two `--no-x` flags are wrong, which is why this reads as correct code at a glance.

## Impact

Two documented flags silently do nothing. The failure is **silent in the worst direction**: an operator who passes `--no-fill` to skip the craft-polish phase gets the full pipeline anyway, with no warning, no error, and no signal in the output that the flag was ignored. FILL is the expensive phase, so the flag exists precisely to avoid work that then happens regardless.

Same class as #1819 and #1345 — a config/CLI key that is live at the declaration site and dead at the read site.

## Suggested fix

Read the keys Commander actually populates:

```ts
if (opts.freshen === false) input.noFreshen = true;
if (opts.fill === false) input.noFill = true;
```

(Or restate the CLI surface in terms of `opts.freshen` / `opts.fill` directly and invert at the `DesignPipelineInput` boundary — either is fine, but the declared flag and the read key must agree.)

Also update the `DesignPipelineCliOptions` interface at the top of the file: it currently declares `noFreshen?: boolean` / `noFill?: boolean`, which is what made the wrong read typecheck cleanly. That type is asserting a shape Commander never produces, so the type system actively concealed this bug rather than catching it.

## Regression test

A test asserting that parsing `['--no-fill']` yields a `DesignPipelineInput` with `noFill: true` (and that parsing without it does not) — i.e. a test at the **option-parsing boundary**, not at the pipeline internals. There is currently no test at that boundary for this command, which is why the inert flags survived.

Note `packages/cli/tests/**` is region-locked by an in-flight PR at time of filing; a colocated `packages/cli/src/commands/design-pipeline.*.test.ts` is an established alternative in this repo.

## Provenance

Found by a `fleet-command` **test-fleet** lane while characterizing behavior of the craft/design commands. The lane correctly **refused to fix code under test** and parked it. Premise independently re-derived at source against `origin/main` `c1ca02ba2` before filing.

The same lane swept **all** craft and design commands for this defect class: `design-pipeline.ts` and `test-craft.ts` (see the sibling issue) are the **only** two instances in the repo.

Sibling: the `--no-source-pair` instance in `test-craft.ts`. Both share one root cause and are a natural single PR.

chadjw:
## Summary

`harness design-pipeline` declared two documented skip-flags, `--no-freshen` and `--no-fill`. **Both were permanently inert** — passing either changed nothing and the phase it names ran anyway.

Root cause is Commander's negated-flag convention applied backwards. `.option('--no-freshen')` defines an option named `freshen` that defaults to `true` and becomes `false` when the flag is passed; it **never** creates a `noFreshen` key. The action read `opts.noFreshen === true` / `opts.noFill === true`, which are `undefined` on every code path, so `input.noFreshen` / `input.noFill` were never set and the engine — which skips a phase only on `=== true` (`design-pipeline/index.ts:55` and `:90`) — always ran FRESHEN and FILL.

The failure was silent in the worst direction. FILL is not only the expensive phase, it is the **only phase that writes into the adopter's repository on a default run** — `phases/fill.ts:98` and `:123` scaffold `design-system/DESIGN.md` and `tokens.json`, and `:108` / `:113` append stub sections onto an existing `DESIGN.md`. So while `--no-fill` was inert there was **no way to stop a bare `harness design-pipeline` from mutating the repo**, with no warning and no signal that the flag had been ignored.

(FIX writes too — `.harness/handoff.json` at `phases/fix.ts:105`, and adopter source files via `runAlignDesignSystem` called without `dryRun` at `phases/fix.ts:43-46` → `align/index.ts:347`. But FIX is gated on `--fix` at `design-pipeline/index.ts:69`, so it is not reachable on the bare run `--no-fill` governs.)

## The fix

```ts
// `=== false` (not falsiness): Commander always populates these, defaulting
// both to `true`, so only an explicit `false` means the flag was passed.
if (opts.freshen === false) input.noFreshen = true;
if (opts.fill === false) input.noFill = true;
```

## Root cause verified empirically, not inferred

Probed the exact Commander in use (12.1.0, resolved via `packages/cli/node_modules/commander`), declaring the two negated options as the real command declares them:

```
[]                           => freshen=true  fill=true   | noFreshen: undefined
["--no-freshen"]             => freshen=false fill=true   | noFreshen: undefined
["--no-fill"]                => freshen=true  fill=false  | noFill:    undefined
["--no-freshen","--no-fill"] => freshen=false fill=false  | both:      undefined
["--no-freshen","--no-freshen"] => freshen=false
["--freshen"]                => THROWS commander.unknownOption
["--no-freshen=false"]       => THROWS commander.unknownOption
after `--` terminator        => freshen=true  (flag becomes an ignored positional — POSIX-correct)
```

Round 2 re-ran this across 15 argv shapes, including adjacency to the variadic `--files`. `freshen` / `fill` are **own properties on every path**, never `undefined`, so the `=== false` read cannot silently no-op. Two things follow:

1. `opts.noFreshen` / `opts.noFill` were `undefined` on **every** path — the old reads were dead code.
2. `freshen` / `fill` are **always present**, defaulted to `true`, even on a bare run. So the fix must test `=== false`. An `=== undefined` test would never fire, and a presence test would set both keys on every run.

## The sweep — corrected twice, now re-derived

`packages/cli/src` declares **15** `.option('--no-…')` options, all under `packages/cli/src/commands/`. **Eleven were never broken; four were ever broken.** (Stating it by defect-class lifetime rather than a snapshot — the snapshot moves as sibling PRs land, and one did mid-review.)

The eleven, split by whether the value crosses a module boundary:

| Site | Read | Shape |
| --- | --- | --- |
| `fleet/budget-check.ts:219` | `noRefresh: opts.refresh === false,` | normalize → engine |
| `recommend.ts:138` | `noCache: opts.cache === false,` | normalize → engine |
| `predict.ts:177` | `noRoadmap: opts.roadmap === false,` | normalize → engine |
| `agent/review.ts:179` | `noMechanical: opts.mechanical === false,` | normalize → engine |
| `roadmap/sync.ts:154` | `allowCreate: options.create !== false,` | normalize → local opts |
| `roadmap/sync.ts:155` | `syncIssueState: options.stateChange !== false,` | normalize → local opts |
| `install.ts:459` | `if (opts.generate === false) return;` | read positive, stays local |
| `adoption.ts:370` | `if (opts.write === false) {` | read positive, stays local |
| `distortion.ts:155` | `if (opts.write === false) {` | read positive, stays local |
| `mcp.ts:243` | `includeSkills: opts.skills !== false,` | read positive, stays local |
| `fix-drift.ts:48` | `const dryRun = options.dryRun !== false;` | read positive, stays local |

The four ever-broken:

| Site | Issue | Status |
| --- | --- | --- |
| `design-pipeline.ts` × 2 | #1881 | **fixed by this PR** |
| `test-craft.ts` | #1882 | fixed by **PR #1954**, merged mid-review — now reads `opts.sourcePair === false` at `:96`, issue closed |
| `dashboard.ts:121` | #1956 | **still broken, still open** |

So on the current base the live count is **12 correct / 3 broken**, and once this PR lands **`dashboard.ts:121` is the only remaining instance of the class.**

The split is not a style preference — it is a clean rule: **normalize when the value crosses into an engine, read the positive key when it stays in the command file.** This PR crosses into an engine (`runDesignPipeline`), so it normalizes. `predict.ts`, `agent/review.ts`, and `roadmap/sync.ts` are its structural twins.

> **Two corrections to this section — please read.**
>
> **(1) The original claim.** Round 1 of this PR listed `fleet/budget-check.ts:170` as a **third broken sibling**, adding that `:219` "already uses the correct form in the *same file* as its broken `:170` read". **False, withdrawn.** `:219` is `:170`'s **producer**: the `.action()` at `:211-221` normalizes `noRefresh: opts.refresh === false` — under the comment *"Commander maps `--no-refresh` to `refresh: false`; normalize to noRefresh"* — and hands the normalized object to the separately-exported `runBudgetCheck()` at `:165`, whose `if (!opts.noRefresh)` at `:170` consumes it. `runBudgetCheck` is also called directly from unit tests with an explicit `noRefresh: true`, so "fixing" `:170` would have broken them.
>
> **(2) The replacement claim was also wrong.** The correction written for it said "exactly **13** declarations, **nine** correct" and attributed the enumeration to issue #1956 "rather than re-derived". Both halves fail. The count came from a single-line grep (`grep -rn "option('--no-"`), which structurally cannot see a **multi-line** `.option(` call — and two exist, `roadmap/sync.ts:175` and `:180`. Re-derived with `grep -rn -- "'--no-" packages/cli/src/`, which sees both forms, the total is **15 / 11 correct / 4 broken**. And #1956 contains **no such table**; its only sweep sentence describes a three-site scan from PR #1954. Citing a document for content it does not contain is the same failure as the original claim, one layer up. The pointer is withdrawn and the enumeration is inlined above and in `provenance.json`.
>
> **Nothing in the shipped diff ever depended on any of these claims** — they were commentary in the scope note, and the broken count of four is unchanged throughout. The code is unaffected.

## Assumptions made

1. **Kept the engine contract as-is.** The engine's `noFreshen` / `noFill` skip checks are untouched. Beyond "the defect is at the command boundary", the decisive constraint is that **`noFreshen` and `noFill` are fields of the published MCP `inputSchema`** for `run_design_pipeline` (`mcp/tools/design-pipeline.ts:32-33`). Inverting the engine contract would break an external tool surface with unknown clients, or force a translation layer that reintroduces the same double negative at a second boundary. The chosen seam confines the negation to one translation point.
2. **Used `opts.freshen === false`, not a truthiness check.** Justified by the 15-shape probe above. Also preserves the command's contract of **omitting** unsupplied flags rather than sending `false` for them.
3. **Also corrected the `DesignPipelineCliOptions` interface** (`noFreshen?`/`noFill?` → `freshen`/`fill`, `design-pipeline.ts:43-44`). Same root cause, not scope creep: the interface asserted a shape Commander never produces, which is *exactly why* the dead reads typechecked cleanly. Round 2 also dropped the `?`, so the type mirrors runtime reality. **Correction:** an earlier round-2 revision claimed this *closed* the hole — that declaring them `?` would make an `=== undefined` read compile where required-ness would not. **It does not.** TypeScript exempts `=== undefined` from its no-overlap check (TS2367), so `opts.freshen === undefined` compiles clean against `boolean` exactly as against `boolean | undefined`; and Commander's `.action()` is typed `(...args: any[])`, so the annotation is never structurally checked anyway. The change is honest documentation with **zero compiler enforcement** — the mirrored test pair is what actually guards the dead read. The JSDoc now says so.
4. **INVERTED AN EXISTING PASSING TEST — please read this one.** At the **pre-fix revision** (`c7bdb08c6~1`), `design-pipeline-command.test.ts:140-172` was a deliberate CHARACTERIZATION block pinning the *broken* behavior. Those three assertions now assert the opposite, so a previously-passing assertion changed meaning. Intended, and anticipated by the original author: *"These tests will go red the moment it lands, which is exactly the signal wanted."* The replacement block is `:140-205`.
5. **Added guard tests beyond a straight inversion.** A bare run must still omit both keys, and each flag alone must not set the other's key. **Round 2 added the missing mirror half** (`:201-204`). A mutation probe confirmed the asymmetry was load-bearing: mutating the `fill` branch to also set `input.noFreshen` **survived** the round-1 suite (46 passed) and **fails** the round-2 suite (1 failed | 46 passed of 47, at `:203`).
6. **Held scope to `design-pipeline.ts` only.** Two sibling instances deliberately not touched: `test-craft.ts` (#1882 — a parallel lane, whose PR #1954 has since merged, closing it) and `dashboard.ts:121` (#1956 — still open).

## Adjacent findings, filed rather than fixed here

This PR is scoped `Closes #1881`. Three findings were filed as follow-ups instead of growing it:

- **#1958** — `printPipelineResult` (`design-pipeline.ts:112-156`, render at `:121-127`) prints `newContext()`'s untouched `false` defaults *as measurements* when FRESHEN is skipped: `DESIGN.md=no tokens.json=no` for a project where the files exist. `--json` serializes the same values with no skip marker.
- **#1960** — with FILL skipped, `anyBootstrap` (`report.ts:46`) stays false and `computeVerdict` (`report.ts:56-64`) returns `pass` for a design-system-less project that reports `warn` without the flag. A false-green verdict; **not** an exit-code regression, since `pass` and `warn` both map to 0.
- **#1959** — `--ci` is a separate, pre-existing inert flag on this same command: declared `design-pipeline.ts:61`, written `:80`, accepted at `design-pipeline/index.ts:35`, read by **nothing** (`FixInput` at `phases/fix.ts:21-27` has no `ci` field). The green test at `design-pipeline-command.test.ts:104-107` reads as coverage while proving only that a value reaches a field with no readers.

> **Correction:** an earlier revision said #1958 and #1960 were "reachable only *because* this fix works". That overstates it. Both states have been reachable since `run_design_pipeline` shipped — its `inputSchema` declares `noFreshen`/`noFill` (`mcp/tools/design-pipeline.ts:32-33`) and passes them straight to the engine at `:71`, and `tests/design-pipeline/integration/end-to-end.test.ts:89-93` and `:95-102` already exercise both. This fix makes them reachable from the **CLI** as well. They are pre-existing defects on a live surface, not regressions introduced here — which *strengthens* the case for filing them separately. Further detail from round-2 review has been added as comments on each issue.

## Reproducing test

`packages/cli/tests/commands/design-pipeline-command.test.ts`, written **before** the fix and observed red, then turned green by the fix with no test edits in between:

- **Before fix:** `3 failed | 43 passed (46)` — `AssertionError: expected undefined to be true`
- **After fix:** `47 passed (47)` (46 after round 1, plus the round-2 mirror guard)

## Verification

- Reproducing test: red → green as above
- Mutation probe on the round-2 mirror guard: asymmetric-`fill` mutant survives round 1, fails round 2 at `:203`
- Commander premise re-probed across 15 argv shapes against the real installed 12.1.0
- Local gates at the reviewed tree: `tsc --noEmit` clean, `eslint` clean, `prettier --check` clean, design-pipeline suites 76/76. The three-OS CI legs were green at `8932e148a`; the current head re-merges `origin/main` (`d60a03ead`, carrying PRs #1952 and #1954) so its run is re-verifying from scratch — read the live check list rather than this line
- `generate-docs --check`: **no drift** — flag declarations and help text are unchanged, only the read logic, the options type, and comments
- Every `file:line` in this body and in `provenance.json` re-verified against the tree **after** the final edit, not before it
- Pushed through all pre-push gates with no `--no-verify`

## Route

`bug` → `harness-debugging` pipeline. Provenance at `docs/changes/1881-design-pipeline-negated-flags/provenance.json`.

Closes #1881


chadjw:
## Review round 2 — what changed

The round-1 REQUEST CHANGES is addressed. **No behavior change**: the four shipped lines in `design-pipeline.ts` are the same as round 1. Everything below is a false-claim withdrawal, a test, a type annotation, or a comment.

### The blocking finding: a verified-false claim, corrected in both places

Round 1 named **`fleet/budget-check.ts:170`** as a third broken sibling of this defect class, and added that `:219` "already uses the correct form in the *same file* as its broken `:170` read."

**False. Withdrawn from `provenance.json` and the PR body.** `:219` is `:170`'s **producer** — the `.action()` at `:211-221` normalizes `noRefresh: opts.refresh === false` and hands the object to the separately-exported `runBudgetCheck()` at `:165`, whose `if (!opts.noRefresh)` at `:170` consumes it. `runBudgetCheck` is also called directly from unit tests with `noRefresh: true`, so acting on the claim would have **broken** them. The accurate siblings are kept.

### Then the correction was itself wrong, twice. Both are now fixed and recorded.

Auditing the fix surfaced a second and a third wave — worth reading, because they are the same failure mode one layer up each time:

1. **"exactly 13 declarations, 9 correct" → actually 15 / 11 never-broken / 4 ever-broken.** The 13 came from a single-line grep (`grep -rn "option('--no-"`) which structurally cannot see a **multi-line** `.option(` call. Two exist — `roadmap/sync.ts:175` and `:180` — both read correctly at `:154-155`. Re-derived with `grep -rn -- "'--no-"`.
2. **The enumeration was cited to #1956 "rather than re-derived". #1956 has no such table** — its only sweep sentence describes a three-site scan from PR #1954. Pointer withdrawn; the full 15-row enumeration is now inlined here and in `provenance.json`.
3. **The `?`-removal JSDoc claimed a compiler guarantee that does not exist.** TypeScript exempts `=== undefined` from its no-overlap check, so `opts.freshen === undefined` compiles clean against `boolean` exactly as against `boolean | undefined` (verified with `tsc --strict`); and `.action()` is typed `(...args: any[])`, so the annotation is never structurally checked. The change is honest documentation with **zero enforcement** — the mirrored test pair is the actual guard. Comment rewritten to say that.
4. **"FILL is the only phase that writes into your repository" → qualified to "on a default run".** FIX writes too (`phases/fix.ts:105`, and source files via `align/index.ts:347`), but is gated on `--fix`, so the downstream claim — that a bare run had no write-suppressor — stands.
5. **"#1958/#1960 are reachable only because this fix works" → withdrawn.** `run_design_pipeline`'s `inputSchema` has declared `noFreshen`/`noFill` since it shipped (`mcp/tools/design-pipeline.ts:32-33`, passed through at `:71`), and `end-to-end.test.ts:89-93`/`:95-102` already exercise both. They are pre-existing defects on a live surface — which *strengthens* the case for filing them separately.
6. Four range/count off-by-ones fixed: `:140-205` (not `:139-205`), `fix.ts:21-27` (not `:21-26`), `end-to-end.test.ts:89-93` + `:95-102` (not `:89-103`), and the round-1 suite count of **46** (not 47).

### Suggestions applied

- **Mirror guard test** the mutation probe found missing (`design-pipeline-command.test.ts:201-204`). Verified load-bearing: mutating the `fill` branch to also set `input.noFreshen` **survives** the round-1 suite and **fails** the round-2 suite at `:203`.
- **`freshen` / `fill` made non-optional** (`:43-44`) — see caveat 3 above about what that does and does not buy.
- **Changeset now carries the `(#1881)` ref** the repo's changelog convention wants.

### Filed rather than fixed here (scope is `Closes #1881`)

#1958, #1959, #1960 — each with a mechanism trace. Round-2 review found errors in **my own filings** too, corrected as comments on each: #1958's root cause was wrong (`context.inputs` has a second writer in FILL, so its "always probe" remedy would not work), it missed `graphAvailable` and the `verifiersRun` shrink, and the `--no-freshen --no-fill` compound state — which makes a fully-configured and a completely empty project render **byte-identically**, both `✓ pass` — was unnamed. #1956 also got a correction (its `.action()` cite is `:134`, not `:136`) plus the 15-row table it was being credited with.

### Verification

All 14 CI checks pass on `55a692a16`, including all three OS legs. Every `file:line` in the body, `provenance.json`, and the four issue comments was re-verified against the tree **after** the final edit — 118 citations in the last audit pass alone. Re-merged against `origin/main` (`d60a03ead`, carrying #1952 and #1954); no `--no-verify` anywhere.

Note #1954 merged mid-review and fixed the `test-craft.ts` sibling, so the sweep is now stated by **defect-class lifetime** rather than a moving snapshot. Live count on this base: 12 correct / 3 broken. **Once this lands, `dashboard.ts:121` (#1956) is the only remaining instance of the class.**

PR #1954, opened 2026-09-06
fix(cli): wire test-craft --no-source-pair to the key Commander populates (#1882)

### LINKED ISSUE #1882: test-craft: --no-source-pair is inert — same Commander negated-flag inversion as the design-pipeline sibling

chadjw:
## Summary

`harness test-craft --no-source-pair` documented "Skip source-pairing resolution" and **did nothing**. Source pairing ran on every invocation, and the flag reported nothing about being ignored. The documented escape hatch for repos where pairing is slow, noisy, or resolves the wrong source file simply did not exist.

Route: `bug` → `harness-debugging` pipeline. Provenance: `docs/changes/1882-test-craft-no-source-pair/provenance.json`.

## Root cause

Commander stores a `--no-x` flag under its **positive** camelCase key — `sourcePair`, `true` by default and `false` when the flag is passed. It never creates a `noSourcePair` key.

`buildInput` (`packages/cli/src/commands/test-craft.ts:86`) read:

```ts
if (opts.noSourcePair === true) input.sourcePair = false;
```

`opts.noSourcePair` is `undefined` on every path, so the branch was unreachable, `input.sourcePair` was never set, and the engine's guard (`input.sourcePair !== false`, `packages/cli/src/test-craft/index.ts:153` and `:198`) always resolved to enabled.

Probed directly against the installed commander 12.1.0:

```
with --no-source-pair : {"sourcePair":false}   (noSourcePair: undefined)
without the flag      : {"sourcePair":true}
```

`TestCraftCliOptions` declared the same non-existent `noSourcePair?: boolean` field. That is **why the unreachable read typechecked cleanly and survived review** — the type declaration was part of the defect, not incidental to it.

Not a regression: `git log` shows the defect present since the command was introduced in `b8d97b0af`. The flag has never worked.

## Fix

Command layer only, two lines of behavior:

- `TestCraftCliOptions.noSourcePair?: boolean` → `sourcePair?: boolean`, so the interface describes the object Commander actually produces.
- `if (opts.noSourcePair === true)` → `if (opts.sourcePair === false)`.

This matches the idiom already used across the repo for the same construct — `--no-generate` (`install.ts:459`), `--no-skills` (`mcp.ts:243`), `--no-write` (`adoption.ts:370`), `--no-cache` (`recommend.ts:138`), `--no-roadmap` (`predict.ts:177`). The engine's `input.sourcePair` contract is **unchanged**.

## Why the defect survived

`packages/cli/tests/test-craft/integration.test.ts:89` — "`--no-source-pair` disables pairing entirely" — already existed and passed. It calls `runTestCraft({ sourcePair: false })` **directly, bypassing the CLI**. The engine was proven correct; the wiring between the flag and the engine was never tested. The new guard therefore sits at the option-parsing boundary.

## Reproducing test — before/after

`packages/cli/tests/commands/test-craft-command.test.ts`

Before the fix (test authored red-first):

```
× disables source pairing at the engine when --no-source-pair is passed
AssertionError: expected undefined to be false // Object.is equality
 Tests  1 failed | 35 passed (36)
```

After the fix:

```
 Tests  36 passed (36)
```

Revert-and-fail re-confirmed: restoring `opts.noSourcePair === true` reproduced the identical assertion failure; restoring the fix returned to green.

## Assumptions made

Every default taken, stated explicitly:

1. **I INVERTED AN EXISTING PINNING TEST — please read this one.** The file previously carried `'does not reach the engine with --no-source-pair (characterizes a live defect)'`, which asserted `expect(capturedInput()).not.toHaveProperty('sourcePair')` and **passed against the broken code**. Its comment said it was expected to fail "loudly, by design" the moment the flag was wired correctly. It is now `'disables source pairing at the engine when --no-source-pair is passed'` and asserts `sourcePair === false`. **A previously-passing assertion has changed meaning.** This is deliberate, and it is the intended trigger the original author documented — not a silent rewrite.
2. Kept the engine's `input.sourcePair` contract as-is (optional boolean; pairing disabled only on an explicit `false`). The defect is in the command layer's read. Both engine call sites already agree and are correct, so changing the engine would be the larger, riskier edit.
3. Asserted the concrete value `sourcePair === false` rather than mere property presence, because the engine disables pairing only on a literal `false` — a property present with any other value would still leave pairing on, so presence alone would be a weaker guard than the contract requires.
4. Renamed the interface field rather than translating at the boundary. `recommend.ts` and `predict.ts` keep a `noX` field on their own *internal* options type and map from `opts.x === false`; here `TestCraftCliOptions` **is** the Commander opts shape, so the field itself had to be renamed to stay honest.
5. **Did not add a separate default-path test** (corrected in review — an earlier revision of this PR added one and this list claimed it). Commander defaults the negated key to `true`, so a fix that forwarded the default would set `sourcePair: true` on the input — but the pre-existing `'omits unsupplied flags from the input entirely'` test already asserts the **whole** input shape (`Object.keys(input)` equals `['path']`) for that exact invocation, so it catches the over-correction with a strictly stronger assertion. A dedicated `not.toHaveProperty('sourcePair')` test is subsumed by it and passes against the pre-fix code, so it would document a guard it does not provide; it was removed and the default-path intent moved into the surviving test's comment. Verified by injecting the naive over-correction (`if (opts.sourcePair !== undefined) input.sourcePair = opts.sourcePair`): the surviving test failed with ``expected [ 'path', 'sourcePair' ] to deeply equal [ 'path' ]`` and passed again on restore. Guarding on the explicit `false` keeps the default path sending no property at all rather than a redundant `true`.
6. No documentation change. The flag's declaration and help text are untouched, so `pnpm run generate-docs` produced no diff in `docs/reference/cli-commands.md`. The docs were never wrong — the code was.
7. The unrelated `.harness/comprehension/**/_module.md` shard updates in this commit were written by the repo's own pre-commit comprehension hook, not by hand.

## Scope

Deliberately **not** touched, though they carry the identical Commander inversion:

- `packages/cli/src/commands/design-pipeline.ts:62-63` — issue #1881, separately owned lane running in parallel.
- `packages/cli/src/commands/dashboard.ts:121` — same defect class, genuinely broken (`opts.noOpen` read at `:121`, `--no-open` declared at `:132`, no normalizing `.action()`), outside this batch. Now filed as **#1956**. It was the last *unfiled* instance of the class — `design-pipeline.ts:62-63` above is still a live instance, tracked as **#1881**.

> **Correction (review round 2).** An earlier revision of this section also listed `packages/cli/src/commands/fleet/budget-check.ts:170` as carrying the same defect. **That was wrong.** `budget-check`'s `.action()` at `:219` normalizes `noRefresh: opts.refresh === false` — carrying the comment *"Commander maps `--no-refresh` to `refresh: false`; normalize to noRefresh"* — and feeds the already-normalized object into the separately-exported `runBudgetCheck()` at `:165`, which is where the `opts.noRefresh` read at `:170` lives. `:219` is `:170`'s **producer**, not an unrelated correct example. `budget-check.ts` handles this pattern **correctly**. (Round-3 correction: an earlier wording called it *the one* correct site. It is not unique — a sweep of all 13 negated flags in `packages/cli/src/commands/` finds nine correct sites: four normalize in `.action()` (`recommend.ts:138`, `predict.ts:177`, `agent/review.ts:179`, `budget-check.ts:219`) and five read the positive key directly (`install.ts:459`, `mcp.ts:243`, `adoption.ts:370`, `distortion.ts:155`, `fix-drift.ts:48`). That is nine correct sites *other than* this fix — ten counting `test-craft.ts:96` itself — against three remaining broken ones (`design-pipeline.ts:62`, `:63`, `dashboard.ts:121`), over a complete population of 13 negated-flag declarations; a repo-wide sweep of `packages/` and `apps/` confirms all 13 live under `packages/cli/src/commands/`.) The claim has been removed from this body and from `provenance.json` (both the SCOPE assumption and `notTouched`). The other three scope claims were verified and stand.
- `packages/cli/src/test-craft/index.ts` — engine contract intentionally unchanged.

## Verification

- `packages/cli` `tests/commands/` + `tests/test-craft/`: **171 files, 2316 tests, all pass** (2317 before the redundant test was removed).
- `tsc --noEmit` on `packages/cli`: clean.
- `pnpm run generate-docs`: no drift.
- Full pre-push gate suite passed with no `--no-verify`: changeset present, format:check, coverage ratchet, reference-docs freshness, tool-catalog check.

## Review round 3

The round-2 correction commit was itself reviewed and found to carry three defects of the same class it fixed; all three are corrected in `6ae072944` (documentation only):

1. **A fresh off-by-one citation.** The new `notTouched` entry cited the frameworks cast at `test-craft.ts:88`; expanding the `sourcePair` doc comment in the same commit shifted the file by +1, so it is at `:89`. Fixed in `provenance.json`, this body, and #1957.
2. **An unsupported superlative** — "the one CLI site that handles this pattern correctly" — replaced with the enumerated sweep above.
3. **Contradictory evidence counts** — `afterFix` claimed "36 passed (36)" four lines above `overCorrectionProbe` reporting "35 passed (35)" for the same file, and `wider` claimed 2317. The committed tree is 35 and 2316; all three entries now state the measurement and the round-2 delta explicitly.

## Review round 2

No blocking code defect was found; the fix itself is unchanged. Applied: the `budget-check.ts` scope correction above (three sites — this body, `provenance.json` `assumptions[6]`, `provenance.json` `notTouched[2]`); removal of the redundant default-path test; and `TestCraftCliOptions.sourcePair` made non-optional to match its doc comment, since Commander always populates it.

Filed as follow-ups, out of this PR's `Closes #1882` scope:

- **#1956** — `harness dashboard --no-open` is inert (the genuinely-broken sibling).
- **#1957** — `test-craft --frameworks` accepts any string via an unchecked `as TestFramework[]` cast at `test-craft.ts:89`, so `--frameworks bogus` filters every test out and prints `No tests found to critique.` at exit 0 — an unmeasured run presenting as clean. The MCP surface enum-constrains this input; the CLI does not.

Closes #1882


chadjw:
### End-to-end verification through the built binary

Beyond the unit-level regression guard, the fix was confirmed against the real compiled CLI (`packages/cli/dist/bin/harness.js`) on a two-file fixture (`src/foo.ts` + `src/foo.test.ts`, which pair), with `HARNESS_CRAFT_LLM=mock`:

```
$ harness test-craft --json                    # without the flag
      "sourcePaired": 1,

$ harness test-craft --no-source-pair --json   # with the flag
      "sourcePaired": 0,
```

Before this change both invocations reported `sourcePaired: 1` — the flag made no observable difference. The summary counter now moves, so the behavior is verifiable by a user from the command's own output rather than only by a unit assertion.

Issue #1965, opened 2026-09-07
Two weekly workflows silently failing for months (two distinct causes): Architecture Snapshot blocked by branch ruleset; Holiday Confidence Tracker never writes its ledger


chadjw:
Two scheduled workflows push directly to `main` and are rejected by the branch ruleset on **every** run. Both have been silently red for months, so the ledgers they maintain are stale.

## Evidence

```
remote: error: GH013: Repository rule violations found for refs/heads/main
error: failed to push some refs
##[error]Process completed with exit code 1
```
Most recent: https://github.com/Intense-Visions/harness-engineering/actions/runs/34120963994

| Workflow | Cron | Failing since | Writes |
|---|---|---|---|
| `Architecture Snapshot` (`.github/workflows/snapshot.yml`) | `0 6 * * 1` | **≥ 2026-06-22** — 12+ consecutive weekly failures, no success in visible history | `.harness/arch/timeline.json` |
| `Holiday Confidence Tracker` (`.github/workflows/holiday-confidence-track.yml`) | `30 7 * * 1` | **≥ 2026-08-10** | trend ledger |

## Impact

**The last architecture snapshot actually committed to `main` was `f8d593648 chore: architecture snapshot 2026-04-06` — five months ago.** `.harness/arch/timeline.json` has been frozen since then while the job ran weekly and failed. Any analysis reading architecture *trend* data is reading a five-month-old series and has no signal that it stopped.

The failure is invisible in normal use: it is a scheduled workflow, so nothing surfaces it on a PR, and `main`'s per-commit CI is unaffected and green.

## Root cause

Both jobs do a bare push to the protected branch:

```yaml
# snapshot.yml:26-27
git diff --cached --quiet || git commit -m "chore: architecture snapshot $(date +%Y-%m-%d)"
git push
```

`permissions: contents: write` grants the token write scope, but the **ruleset** (`Main branch protection`, `required_approving_review_count: 1`) applies to `GITHUB_TOKEN` pushes too. Scope is not a bypass.

## Suggested fix — the pattern already works here

The baseline-refresh bot solves exactly this by **opening a PR** rather than pushing (#1937–#1950 all merged normally today). Converting both jobs to `peter-evans/create-pull-request` (or equivalent) makes them consistent with the existing bot and needs no ruleset change.

[...]

PR #1981, opened 2026-09-08
fix(ci): make the Holiday Confidence Tracker's fail-soft contract reachable

chadjw:
## The issue's stated root cause for this workflow is wrong

**#1965 groups `holiday-confidence-track.yml` with `snapshot.yml` under the `GH013` ruleset push
rejection. That is not why this one fails.** It dies two defects earlier, at `git add`, and has never
once reached its own `git push`. From run
[34128383695](https://github.com/Intense-Visions/harness-engineering/actions/runs/34128383695), step
`Commit the trend ledger`:

```
No parseable KPI output; recording nothing (fail-soft).
...
fatal: pathspec '.harness/metrics/holiday-confidence.jsonl' did not match any files
##[error]Process completed with exit code 128
```

`GH013` is real here, but it is the **third** layer, not the first. A reviewer reading #1965 and
applying only the snapshot fix would have watched this workflow fail identically the following
Monday.

**5/5 consecutive failures. It has never succeeded once since inception on 2026-08-10.**
Runs `31372352672`, `32008807443`, `32704972836`, `33407509136`, `34128383695`.

## Four stacked defects

Each was hidden behind the one above it, so each had to be cleared to see the next.

### 1. The workflow's own documented contract took the job red

The header promises: *"this job records NOTHING and still succeeds — a missing data point is honest,
a fabricated one is not."* But when the KPI abstains, no ledger file exists, and the commit step then
runs `git add <that missing path>` under `shell: /usr/bin/bash -e` → exit 128. **The documented
abstain path could never succeed.**

Fixed with an existence guard before staging. Not `continue-on-error:` (that hides a failure) and not
a fabricated data point — this exits clean *because nothing failed*, which is exactly what the header
describes.

### 2. The ledger could never be staged even when written

`.harness/.gitignore` already carried the correct rule, added by this workflow's **own originating
PR** (`0563679fd`, #1131):

```
# metrics/: track the holiday-confidence trend ledger, ignore everything else
metrics/*
!metrics/holiday-confidence.jsonl
```

An unrelated docs commit — `9e2b9b46d` *"docs(roadmap): register 4 deepseek-harness adoption
candidates"* — later re-added a bare `metrics/` at line 46. A trailing-slash directory ignore stops
git descending into the directory at all, so the earlier negation is unreachable regardless of
ordering. The un-ignore was silently switched off by a docs commit, and nobody noticed because the
workflow was already red one defect earlier.

Removed the stray rule and hardened the surviving block's comment against re-introduction. Verified
in **both** directions:

| Path | Before | After |
| --- | --- | --- |
| `.harness/metrics/holiday-confidence.jsonl` | ignored, `git add` refused | **stageable** |
| `.harness/metrics/adoption.jsonl` (CLI telemetry) | ignored | **still ignored** |

### 3. Only now does it reach GH013

Replaced the bare `git push` with this repo's own push-then-fall-back-to-PR idiom, ported in shape
from `ci.yml` (~L326-360, ~L473-496), `release.yml` (~L141-165) and `roadmap-auto-done.yml`
(~L181-203): three direct-push attempts with `rebase --abort` + `reset --hard` between them, then
branch-from-tip → `gh pr create` → `assert-diff-scope.mjs` → self-approve + auto-merge.

Two supporting changes the fallback requires:

- **`pull-requests: write`.** Under the default `read`, `gh pr create` fails with *"Resource not
  accessible by integration (createPullRequest)"* **after** the branch is already pushed — stranding
  the branch and dropping the record. This is the exact failure `roadmap-auto-done.yml:31-36`
  documents having shipped with, where *"the fallback shipped without this widening, so it had never
  once succeeded."*
- **`fetch-depth: 0`**, so the rebase and branch-from-tip have real history.

### 4. Found during the audit — in neither the brief nor #1965

Fixing 1-3 alone would have shipped a workflow that is **green and records nothing, forever.** The
abstain was never honest.

The root program declares `--json` (`packages/cli/src/index.ts:76`) and so does the subcommand.
Commander binds a repeated flag to the **first** declarer, so `harness holiday-confidence --json`
stores `json: true` on the *program* and leaves the subcommand's `opts.json` `undefined`. Probed
against the built CLI:

```
["holiday-confidence","--json"]                => subOpts: {}             progOpts: {"json":true}
["holiday-confidence","--window","7"]          => subOpts: {"window":"7"} progOpts: {}
["holiday-confidence","--window","7","--json"] => subOpts: {"window":"7"} progOpts: {"json":true}
```

`--window` lands on the subcommand; `--json` lands on the program. So `--json` was a silent no-op,
the command rendered pretty text, and the workflow's `JSON.parse` threw on every single run.
`No parseable KPI output` was the symptom of **this**, not of an empty window.

Fixed by reading `cmd.optsWithGlobals()` — the idiom `snapshot.ts:240`, `usage.ts:143`,
`scan-config.ts:176` and `create-skill.ts:250` already use for exactly this collision. Pinned by 4
regression tests that parse through a root program which also declares `--json` (the only
arrangement in which the shadowing reproduces).

**Revert protocol:** reverted to `.action(async (opts) => ...)` → `2 failed | 2 passed`; restored →
`4 passed`.

## Also

- `HUSKY: '0'` on the commit-back step (precedent: `roadmap-auto-done.yml:152`).
- Header rewritten. The contract it documents changed, and a stale header on a file whose entire
  defect was a broken documented contract is its own defect.
- The abstain log now echoes the byte count and first 20 lines of the stdout it could not parse. A
  silent one-liner is exactly how defect 4 survived five runs unnoticed.

## Known-and-not-fixed: the `--json` shadowing is systemic

**36 command files declare `--json`**, and the shadowing applies to all of them:

```
insights  => subOpts.json: undefined   progOpts.json: true
adoption  => subOpts.json: undefined   progOpts.json: true
verify    => subOpts.json: undefined   progOpts.json: true
```

Every one of those that reads `opts.json` directly has a `--json` flag that does nothing. That is a
repo-wide CLI defect with its own blast radius and test burden, **deliberately left out of this PR**
and reported to the fleet orchestrator for routing. Fixed here for the one command this workflow
depends on, and no further.

## Scope note

The brief scoped this lane's diff to `holiday-confidence-track.yml` + `.harness/.gitignore`. Defect 4
extends it to `packages/cli/src/commands/holiday-confidence.ts` plus a new test file. The
justification is the workflow-audit skill's Iron Law — *"a gate that never fires is worse than no
gate; it manufactures false confidence"* — and this fleet's hard gate against hiding a failure:
shipping 1-3 without 4 produces a permanently green, permanently empty tracker. One line of
behaviour, in an existing idiom, colliding with no sibling lane.

`.github/workflows/snapshot.yml` is **not touched** — that is a concurrent sibling lane's diff.

## Verification

This workflow's `on:` is `schedule` + `workflow_dispatch` only, so **PR CI cannot execute it**. PR CI
validates the CLI change and the repo gates. The workflow itself is verified by:

- YAML parses; `permissions` resolves to `{contents: write, pull-requests: write}`; checkout carries `fetch-depth: 0`
- `bash -n` clean on all four `run:` blocks; `node --check` clean on the embedded heredoc
- the commit step's three branches simulated in a scratch git repo:

  | Scenario | Result |
  | --- | --- |
  | ledger absent (abstain) | exits **0** — *was exit 128* |
  | ledger present, new record | commit created, 1 insertion |
  | ledger present, unchanged | exits **0** |

- end-to-end: `harness holiday-confidence --json > result.json` then `JSON.parse` now succeeds
  (it threw on every prior run)
- `4 passed (4)` CLI regression tests; typecheck clean; lint clean; whole-tree `format:check` clean

First live proof is the next Monday run, or a manual `workflow_dispatch` after merge.

## Artifacts

- `docs/changes/holiday-confidence-ledger-commitback/plans/2026-09-07-holiday-confidence-ledger-commitback-plan.md` — full `harness-workflow-audit` trace (INVENTORY → MECHANICAL → JUDGMENT → REPORT), all findings with patches, forks taken, assumptions
- `docs/changes/holiday-confidence-ledger-commitback/audit-session.md` — investigation log
- `docs/changes/holiday-confidence-ledger-commitback/provenance.json` — machine-readable provenance

Refs #1965.


chadjw:
### pr-fleet review-assist note — regression-test coverage is narrower than the description implies

Review verdict on this PR is **clean**; the two central claims were independently reproduced rather than taken on faith:

- **The `--json` shadowing premise is TRUE.** Reproduced byte-for-byte in a standalone Commander repro: the root `--json` (`packages/cli/src/index.ts:76`) is shadowed by the subcommand's own `--json` (`holiday-confidence.ts:74-89`). The fix uses the pre-existing `optsWithGlobals()` idiom already used in `snapshot.ts:240`, `usage.ts:143`, `scan-config.ts:176`, `create-skill.ts:250` — it removes the shadowing rather than masking it.
- **The fail-soft contract was not turned into a no-op.** The existence guard short-circuits only on genuine KPI abstention; real `gh pr create` / `assert-diff-scope.mjs` failures still fail the job under `bash -eo pipefail`. All three scenarios were simulated against a real bare origin.

**The correction worth recording:** of the four regression tests, **only two actually pin the fix.** Verified by revert protocol — reverting `holiday-confidence.ts` to its pre-fix `.action(async (opts) => ...)` signature yields `2 failed | 2 passed`. The other two pass with the bug present and are refactor safety nets, not regression guards.

That is not a blocker, but the PR description reads as though four tests pin the behaviour, and a future reader trusting that count would over-estimate the protection here.

One pre-existing follow-up, not introduced by this PR and not blocking: the ledger sole-writer assumption in the PR-fallback path is unenforced, and is shared with `ci.yml`, `release.yml`, and `roadmap-auto-done.yml`.

This branch was merged forward onto current `main` by `pr-fleet` to clear a false-red conflict (see #2042) and to re-run CI against a fresh base — no source changes were made by the fleet.

Issue #531, opened 2026-06-04
Harden BASELINE_AUTOAPPROVE_PAT self-approval scope


chadjw:
`.github/workflows/ci.yml:158-176` — the refresh-baselines job opens a PR and self-approves using `BASELINE_AUTOAPPROVE_PAT` when branch protection blocks the direct push. Today the auto-approval fires regardless of what's in the PR. Constrain auto-approval to PRs whose diff is _exactly_ `*-baselines.json` and nothing else. Add a defensive check that fails if the PR diff touches anything outside baselines. Source: Pass 1 #8.

chadjw:
## Harness Analysis: harden-baseline-auto-75034915

**Risk:** low (80% confidence)
**Route:** local
**Analyzed:** 2026-06-05T01:33:11.336Z

- Structural complexity: 0.00 (3 files, 0 services, 0 modules affected)
- Semantic complexity: 0.26 (1 unknowns, 1 ambiguities, 1 risk signals)
- Historical complexity: 0.00 (from past execution outcomes)
- Overall: 0.09 → risk level "low", recommended route "local"

<details>
<summary>Full Analysis Data</summary>

```json
{
  "_harness_analysis": true,
  "_version": 1,
  "issueId": "harden-baseline-auto-75034915",
  "identifier": "harden-baseline-auto-75034915",
  "spec": {
    "id": "harden-baseline-auto-75034915",
    "title": "Harden BASELINE_AUTOAPPROVE_PAT self-approval scope",
    "intent": "Constrain the `BASELINE_AUTOAPPROVE_PAT` workflow to only auto-approve pull requests that modify baselines files and nothing else.",
    "summary": "Modify the CI workflow responsible for refreshing baselines. Currently, the automated self-approval triggered by `BASELINE_AUTOAPPROVE_PAT` occurs indiscriminately when branch protection fails a direct push. This needs to be restricted so that the auto-approval only happens if the pull request diff contains changes exclusively within files matching the pattern `*-baselines.json`. A defensive check must also fail the build/PR if any non-baseline files are modified.",
    "affectedSystems": [
      {
        "name": "CI Workflow Engine",
        "graphNodeId": "module:templates/gin",
        "confidence": 0.7,
        "transitiveDeps": [
          "file:templates/gin/main.go"
        ],
        "testCoverage": 0,
        "owner": null
      },
      {
        "name": "GitHub Actions",
        "graphNodeId": "module:packages/core/tests/interaction",
[...]

PR #2048, opened 2026-09-08
fix(ci): bounded retry + clean abstain for the refresh-baselines merge race

chadjw:
## Summary

The `refresh-baselines` job in `.github/workflows/ci.yml` has reddened `main` three times in three days. This makes its final auto-merge call survive a concurrently-advancing base: a bounded retry that re-derives on the fresh tip, then a clean close-and-abstain on exhaustion.

Produced by the `cicd-fleet` remediation lane (item R1), via the `harness-workflow-audit` pipeline. Plan artifact: [`docs/changes/refresh-baselines-merge-race/plans/plan.md`](docs/changes/refresh-baselines-merge-race/plans/plan.md).

### Root cause

This repo has **no `required_status_checks` branch rule at all** — only a ruleset requiring one approving review, and that review is applied inline by the PAT on the line immediately above the merge. `gh pr merge --auto` only *queues* when something is still pending to wait on; with nothing left to wait for it attempts an **immediate** merge. Under a merge burst `main` advances between the earlier `gh pr create` and this call, the `mergePullRequest` GraphQL mutation is rejected, the step exits non-zero under `bash -e`, and **`main` goes red** (which also fires the `Main Health Alarm` workflow) — for a race that loses no work whatsoever.

### Evidence (three CI runs, identical verbatim failure)

| Run | Date | main SHA |
| --- | --- | --- |
| [`34227768312`](https://github.com/Intense-Visions/harness-engineering/actions/runs/34227768312) | 2026-09-08 | `fe2895b5` |
| [`34042737396`](https://github.com/Intense-Visions/harness-engineering/actions/runs/34042737396) | 2026-09-06 | `c83050a3` |
| [`34042651655`](https://github.com/Intense-Visions/harness-engineering/actions/runs/34042651655) | 2026-09-06 | `9eed8eb0` |

```
GraphQL: Base branch was modified. Review and try the merge again. (mergePullRequest)
##[error]Process completed with exit code 1.
```

### The fix

The PR-fallback tail of `Commit refreshed baselines` is restructured into two shell functions plus a bounded retry loop, matching the house style already proven in `.github/workflows/holiday-confidence-track.yml` → `Commit the trend ledger` (`for attempt in 1 2 3` with a re-fetch between attempts), so the repo's two commit-back jobs stay recognisably the same shape.

- **`rebuild_branch_on_main`** — the existing "Fix B" branch construction (fetch `origin/main`, `checkout -B` from its tip, re-apply `$BASELINE_FILES` from `$OURS`, re-apply the transient-allowance deletions, commit), factored out so a retry **re-derives on the newer tip** rather than blindly repeating. Safe by construction: the baseline files are regenerated wholesale, so "ours" is always the correct resolution and the re-apply cannot conflict. Returns non-zero when the new tip already carries these baselines.
- **`approve_pr`** — the unchanged `#531` fail-closed scope guard (`assert-baseline-only-diff.mjs`) followed by the PAT self-approval, re-run after each force-push because a force-push can dismiss the inline approval.
- **Retry loop** — `for attempt in 1 2 3` with `sleep 5` between attempts. On failure: re-derive, `git push --force-with-lease` **the same branch** (same PR, same URL — never a second PR), re-approve, retry. `--auto` is kept so the job still behaves correctly if required checks are ever added.

### F1 decision (human, taken as given): close the superseded PR on exhaustion

The one design fork was what happens to the refresh PR when the retries are exhausted: **(a)** close it and exit clean, or **(b)** leave it open for a human. The human chose **(a)**.

Rationale, preserved in the workflow comment: `.harness/**/baselines.json` uses a **`merge=ours`** custom merge driver, so a *stale* baseline PR resolves in the PR's favour — landing it later would **REVERT** `main`'s newer baselines. A superseded baseline-refresh PR is therefore *actively dangerous*, not merely stale, and leaving it open to accumulate is the wrong answer. Closing is correct because the very next merge to `main` regenerates a fresh one.

On exhaustion (or when a re-derivation finds `main` already carries these baselines, or the force-push loses its lease) the job emits a `::notice::`, closes the PR with an explanatory comment via `gh pr close --delete-branch`, and exits 0. **This is a visible abstain, not a swallowed failure** — nothing is lost and the outcome is stated in the run log and on the closed PR, the same posture `holiday-confidence-track.yml` already takes when its KPI abstains.

### Explicitly NOT done

- No `|| true` appended to the merge.
- The `assert-baseline-only-diff.mjs` scope guard (#531) is unchanged, still fails closed, and is now re-run before *every* approval rather than once.
- The PAT self-approval guard is unchanged.
- The deliberate per-commit `push` concurrency group (`ci.yml:31`) is untouched.

## Regression guard

Three assertions added to `tests/scripts/baseline-gating.test.mjs` — already the home of the `#671` refresh-race regression tests, and already CI-wired as the `Baseline-gating regression test` step (`node --test 'tests/scripts/*.test.mjs'`). They assert over the workflow text, since the defect lives in the YAML rather than in a script:

1. the `gh pr merge` call sits inside a bounded `for attempt in 1 2 3` retry that re-derives via `rebuild_branch_on_main`;
2. exhaustion closes the superseded PR (`gh pr close --delete-branch`), announces it with `::notice::`, and exits 0;
3. the merge is not neutralised with `|| true` and the `#531` scope guard still runs.

All three **fail against the pre-fix workflow** (verified against `git show HEAD~1:.github/workflows/ci.yml`), so they are not vacuous. Full file: 19/19 pass under Node 22.

## Also in this diff

Four `.harness/comprehension/**/_module.md` shards were regenerated by the repo's own pre-commit comprehension hook (source-hash refreshes; three of them are pre-existing drift on `main` that the hook picks up on any commit). Not hand-edited.

## Adjacent findings, filed as follow-ups rather than widening this diff

- A **job-level** `concurrency:` group on `refresh-baselines` (`cancel-in-progress: false`) would serialize the refresh runs and shrink the race window this retry loop absorbs, without touching the deliberate per-commit verification concurrency. It would not fully close the race (`main` also advances from ordinary PR merges), so it complements rather than replaces this fix.
- Whether the repo *should* carry a `required_status_checks` rule at all — its absence is precisely what turns `--auto` into an immediate merge here, and that has implications well beyond this job.

---

**Do not auto-merge** — this branch is handed to a human to land.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -333,30 +333,71 @@ jobs:
             exit 0
           fi
           echo "Direct push blocked by branch protection. Creating PR instead."
+          BRANCH="chore/refresh-baselines-$(date +%s)"
+
           # Fix B: base the PR branch on the CURRENT tip of origin/main, not the
           # (possibly stale) event SHA this run checked out. Concurrent refresh runs
           # otherwise branch from divergent commits and conflict on the same baseline
           # lines; GitHub auto-merge cannot resolve content conflicts, so the loser
           # sits open forever (this was the #671 failure mode).
-          git fetch origin main
-          BRANCH="chore/refresh-baselines-$(date +%s)"
-          git checkout -B "$BRANCH" origin/main
-          # Re-apply our freshly generated baselines on top of latest main. The files
-          # are regenerated wholesale, so "ours" is always the correct resolution and
-          # this re-apply can never conflict.
-          git checkout "$OURS" -- $BASELINE_FILES
-          git add $BASELINE_FILES
-          # Re-apply the transient-allowance cleanup on top of latest main (the merge may
-          # have carried allowance files onto main). --ignore-unmatch keeps this a no-op
-          # when none are present.
-          for d in $ALLOWANCE_DIRS; do
-            git rm -q -r --ignore-unmatch "$d" >/dev/null 2>&1 || true
-          done
-          if git diff --cached --quiet; then
+          #
+          # Factored into a function because the merge retry below re-derives on the
+          # NEW tip between attempts rather than blindly repeating a merge that main
+          # has already moved out from under. Returns non-zero when the latest main
+          # already carries these baselines (nothing left to propose).
+          rebuild_branch_on_main() {
+            git fetch origin main
+            git checkout -B "$BRANCH" origin/main
+            # Re-apply our freshly generated baselines on top of latest main. The files
+            # are regenerated wholesale, so "ours" is always the correct resolution and
+            # this re-apply can never conflict.
+            git checkout "$OURS" -- $BASELINE_FILES
+            git add $BASELINE_FILES
+            # Re-apply the transient-allowance cleanup on top of latest main (the merge may
+            # have carried allowance files onto main). --ignore-unmatch keeps this a no-op
+            # when none are present.
+            for d in $ALLOWANCE_DIRS; do
+              git rm -q -r --ignore-unmatch "$d" >/dev/null 2>&1 || true
+            done
+            if git diff --cached --quiet; then
+              return 1
+            fi
+            git commit -m "chore: refresh baselines after merge [skip ci]"
+          }
+
+          # Scope guard + self-approval. Re-run after every force-push, because a
+          # force-push can dismiss the inline approval applied to the previous head.
+          approve_pr() {
+            # Defensive scope check (#531): the branch is constructed to contain only
+            # baseline files, but never self-approve a PR whose diff reaches beyond
+            # them. $BASELINE_FILES is the single source of truth for the allowlist.
+            # Fails closed — any unexpected path (or an empty diff) aborts before approval.
+            gh pr diff "$PR_URL" --name-only | node scripts/assert-baseline-only-diff.mjs $SCOPE_ALLOW
+            # Approve inline with the PAT. Events triggered by GITHUB_TOKEN do not start
+            # workflow runs, so a separate auto-approve workflow on pull_request_target
+            # would never fire for this PR.
+            GH_TOKEN="$AUTOAPPROVE_PAT" gh pr review "$PR_URL" --approve \
+              --body "Auto-approved: baseline refresh from github-actions[bot]."
+          }
+
+          # Close the PR and abstain cleanly. `.harness/**/baselines.json` uses a
+          # `merge=ours` custom merge driver, so a STALE baseline PR resolves in the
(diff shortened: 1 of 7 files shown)


chadjw:
## cicd-fleet VERIFY note (independent — not the lane's self-report)

**Verdict: verified, with one caveat a reviewer should know about.**

### What the green does *not* cover

The `refresh-baselines` job this PR modifies is gated:

```yaml
if: github.ref == 'refs/heads/main' && github.event_name == 'push'
```

so it **skips on pull requests** — and indeed reports `skipping` on this PR. **The 12 green checks here do not exercise the changed code at all.** Treating this PR's green as proof the fix works would be exactly the "reported success without running" trap. It is not.

### What actually covers the change at PR time

`tests/scripts/baseline-gating.test.mjs`, run by the ubuntu `Baseline-gating regression test` step (`node --test 'tests/scripts/*.test.mjs'`). All three new assertions executed and passed in run `34244254449`:

```
ok 29 - refresh-baselines: the auto-merge call sits inside a bounded retry
ok 30 - refresh-baselines: retry exhaustion closes the superseded PR and exits clean
ok 31 - refresh-baselines: the merge is not neutralised and the #531 scope guard survives
```

That verifies the **shape** of the fix (retry present, abstain path present, merge not neutralised, `assert-baseline-only-diff.mjs` guard intact). It does not and cannot verify **runtime** behaviour against a live merge race.

**The first post-merge run on `main` is the real confirming signal.** If it goes red on the merge step again, the retry is not doing its job.

### Independently confirmed

- Plan artifact present at `docs/changes/refresh-baselines-merge-race/plans/plan.md`, tracing a real `harness-workflow-audit` run.
- Failure is **removed, not hidden**: no `|| true` on the merge; the #531 scope guard and PAT self-approval both survive, and test 31 guards against a future change that neutralises either.
- Retry genuinely **re-derives** on the new tip between attempts rather than blindly repeating.
- `approve_pr` is correctly re-run after each force-push (a force-push dismisses the prior inline approval) and uses `--force-with-lease`.
- All 3 OSes green: ubuntu, macOS, windows.

### Base freshness

CI was gathered against pinned base `db9c6739da57f6dd38a2e9d83bf94a1bfce9ca7b`; `main` has since advanced to `fef03ac5e`. This repository has **no `required_status_checks` rule**, so nothing enforces up-to-date-before-merge and this green is **provisional against a moving main** — re-check before landing.

*`cicd-fleet` never merges. This PR is handed back unmerged for human review.*

Issue #1867, opened 2026-09-06
Persona CI workflow generator emits the per-ref cancelling concurrency group on push:main — 4 generated workflows inherit #1865's defect


chadjw:
## Summary

Four **generated** persona CI workflows carry the same concurrency-supersession defect that PR #1865 fixes in `ci.yml` and `harness.yml`: they trigger on `push: branches: [main]` while using a per-**ref** concurrency group with `cancel-in-progress: true`. On a push to `main`, `github.ref` is `refs/heads/main` for every commit, so all main pushes collapse into one group and each new merge cancels the still-running run of the previous commit.

Filed by `cicd-fleet` as the explicit follow-up to PR #1865, which deliberately did **not** fold these in — they are generated files guarded by `pnpm generate:persona-workflows:check`, so hand-editing them would be reverted by the drift guard. **The fix belongs in the generator.**

## Root cause — one source, four outputs

`packages/cli/src/persona/generators/ci-workflow.ts:197-204`:

```ts
if (runner === 'workspace') {
  // Cancel superseded runs on rapid pushes so advisory jobs don't pile up
  // (matches harness.yml / pr-advisory-checks.yml).
  workflow.concurrency = {
    group: '${{ github.workflow }}-${{ github.ref }}',
    'cancel-in-progress': true,
  };
}
```

Note the comment cites `harness.yml` as the pattern being matched — and `harness.yml` is precisely one of the two files PR #1865 corrects. The generator is propagating the antipattern.

## Affected outputs (verified)

Persona workflows that trigger on `push: [main]` **and** carry the per-ref cancelling group:

| Workflow | push branches |
| --- | --- |
| `.github/workflows/persona-architecture-enforcer.yml` | `main`, `develop` |
| `.github/workflows/persona-documentation-maintainer.yml` | `main` |
| `.github/workflows/persona-graph-maintainer.yml` | `main` |
| `.github/workflows/persona-task-executor.yml` | `main`, `develop` |

Not affected (no `push` trigger, so per-ref cancellation is correct): `persona-codebase-health-analyst.yml`, `persona-entropy-cleaner.yml`, `persona-performance-guardian.yml`.

## Severity

Lower than #1865's: these jobs are **advisory** (`warn-not-fail`) and `permissions: contents: read`, so a cancelled run does not gate a merge. The cost is loss of the advisory signal on most main commits, in exactly the merge bursts where drift is most likely to be introduced.

## Proposed fix

[...]

PR #2049, opened 2026-09-08
fix(persona): stop generated persona workflows cancelling main's verification (#1867)

chadjw:
## Summary

The persona CI-workflow **generator** emits a per-ref concurrency group with an unconditional `cancel-in-progress: true`. Four of the seven generated workflows also trigger on `push:` to a trunk branch, where `github.ref` is `refs/heads/main` for **every** commit — so all main pushes share one concurrency group and each merge **cancels the still-running verification of the previous commit**.

The run concludes `cancelled`, not `failure`, so **nothing alarms**. The verdict is destroyed silently.

This is #1865's defect in its third home. #1865 fixed the hand-written `ci.yml` / `harness.yml` and explicitly deferred the generated set (its follow-up 1: *"Same concurrency defect in the 4 generated persona workflows → fix via their generator"*). #1867 is that follow-up.

Refs #1867. Precedent: #1865 (merge commit `93bc0788f`).

## Evidence

Live, at the pinned base `db9c6739d`. On the `75eade4ee` main push, three persona runs were cancelled when `2f1e5bd04` landed four minutes later:

| workflow | run id | conclusion |
| --- | --- | --- |
| Documentation Maintainer | `34239670105` | `cancelled` |
| Graph Maintainer | `34239670102` | `cancelled` |
| Task Executor | `34239670064` | `cancelled` |

**The repo already documents the symptom.** `scripts/main-health-check.mjs:32` — *"`cancel-in-progress: true`, so rapid merges leave a trail of cancelled runs"* — and its `DECISIVE_CONCLUSIONS` has to exclude `cancelled` to work around it.

### Which of the seven are actually affected

All 7 carry the defective block; **4** have a `push` trigger and are therefore hit by the trunk-cancellation defect. Correcting the brief I was given (which said 7 trigger on push) — #1867's own title says 4, and 4 is right:

| generated workflow | `push` | `pull_request` | `schedule` | hit by the defect |
| --- | --- | --- | --- | --- |
| `persona-architecture-enforcer.yml` | yes (`main`, `develop`) | yes | yes | **yes** |
| `persona-documentation-maintainer.yml` | yes (`main`) | yes | no | **yes** |
| `persona-graph-maintainer.yml` | yes (`main`) | no | yes | **yes** |
| `persona-task-executor.yml` | yes (`main`, `develop`) | yes | no | **yes** |
| `persona-codebase-health-analyst.yml` | no | yes | yes | no |
| `persona-entropy-cleaner.yml` | no | no | yes | no |
| `persona-performance-guardian.yml` | no | yes | yes | no |

## The fix

Applied at `packages/cli/src/persona/generators/ci-workflow.ts` — #1865's blessed shape, verbatim, so all seven regenerate with it:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

- **push → per-COMMIT group (`github.sha`), never cancelled** — every commit landing on a trunk branch reaches its own persona verdict.
- **PR → per-REF group, cancel-in-progress ON** — a new push to a PR still supersedes the old run. **PR runner spend is unchanged.**

The emitted block carries a comment citing the three cancelled run IDs and both issues, so the next reader does not "simplify" it back — matching how #1865 annotated `ci.yml` / `harness.yml`.

Consistency with the already-blessed fix was the goal, not a new approach. #1865 had already validated that `cancel-in-progress` officially accepts an expression and that the `A && B || C` idiom runs on this repo's runners.

## Tests

Two regression tests in `packages/cli/tests/persona/generators/ci-workflow.test.ts`:

1. `does not emit a cancelling concurrency group for the push-to-trunk path (#1867)` — asserts `cancel-in-progress` is never a literal `true` and the group keys on `github.sha`.
2. `still cancels superseded runs on the pull_request path (#1867)` — guards against over-correcting into a PR runner-spend regression.

**Both were verified to fail against the pre-fix generator** (stashed the fix, ran `-t 1867` → `2 failed`; restored → `20 passed`). A generator defect that regenerates silently is exactly what let this survive #1865, so a test that cannot fail would have been worthless here.

One pre-existing assertion (`expect(workflow.concurrency['cancel-in-progress']).toBe(true)`) encoded the defect as expected behaviour and was updated.

## Remediation actions / assumptions made

1. **Root cause: the generator, not the seven files.** The `.yml` files are generated and guarded by `pnpm generate:persona-workflows:check` in CI. Hand-patching them would have been reverted by the next regeneration.

2. **F2 — the human's binding decision: fix the generator only.** `packages/cli/src/commands/ci/init.ts:107` is a **second, adopter-facing emitter carrying the identical defect** (this repo's known dual-source-of-truth hazard). Option (b), fixing it too, was **explicitly not taken, on blast-radius grounds**: it changes scaffolding output for every adopter, which is materially larger than a change to this repo's own dogfooded workflows. It is a real defect, so it is **filed as a follow-up, not dropped** — see below.

3. **Applied the shape unconditionally rather than only when a `push` trigger is present.** Keeps generated output uniform and identical to #1865. Consequence for the three non-`push` workflows: on their `schedule` path a new run now queues behind an in-flight one instead of cancelling it. These are weekly/daily crons on advisory single-job workflows — immaterial, and the same trade #1865 accepted.

4. **Runner spend on trunk rises with burst size — stated plainly.** That increase *is* the fix: verifying every commit necessarily costs more than verifying one in four. These are single-job `ubuntu-latest` advisory workflows, so the absolute cost is far below #1865's three-OS matrix.

5. **Did not touch `scripts/main-health-check.mjs`.** Its comment goes partially stale, but its logic stays correct (excluding `cancelled` is still right; there will just be far fewer). Out of scope — already a #1865 follow-up.

6. **Five `_module.md` comprehension shards are in the diff.** These are written by the repo's pre-commit comprehension hook, not by hand; three cover directories this change does not touch (the known local-only driver behaviour). Called out so they are not mistaken for scope creep.

## Second finding from the same audit — filed, not fixed here

The `harness-workflow-audit` run surfaced an **independent error-severity defect**, reported rather than folded in:

**`persona-architecture-enforcer.yml:14` — `paths: src/**` resolves to 0 tracked files.** There is no top-level `src/` in this repo (code lives under `packages/*/src/`); `git ls-files src/` → **0**. The Architecture Enforcer `pull_request` gate has therefore **never fired**. Three siblings carry the same dead glob but survive on a second one (`docs/**` → 2082 files, `packages/**` → 4332).

Out of scope: those globs come from `agents/personas/*.yaml`, not from the concurrency logic. Filed separately. Per the audit skill's escalation rule, re-arming a long-dead gate should be preceded by one manual run against the current tree — it will surface a backlog.

## Verification

- `pnpm generate:persona-workflows:check` (the CI drift guard) → **`OK — 7 persona workflows are up to date`**
- All 7 regenerated files re-parse as YAML; programmatic assert that none retains `cancel-in-progress: true` and all key on `github.sha`
- `grep` for the defective group/flag across the 7 → **zero residual occurrences**
- `ci-workflow.test.ts` → **20 passed** (was 18)
- `tsc --noEmit` on `@harness-engineering/cli` → clean
- `pnpm format:check` (whole tree) → all files match Prettier style
- Committed and pushed through **all** pre-commit and pre-push gates with **no `--no-verify`** (arch gate, traceability, lint-staged, comprehension compile, changeset check, `generate-docs --check`, tool-catalog check)
- Explicitly **not** touched: `packages/cli/src/commands/ci/init.ts`, `agents/personas/*.yaml`, `scripts/main-health-check.mjs`

## Plan artifact

`docs/changes/persona-workflow-cancel-1867/plans/2026-09-08-persona-workflow-cancel-plan.md` — full audit output (inventory / mechanical / judgment / report), the clean checks, the F2 decision record, 10 observable truths with their gates, and the change delta.

## Follow-ups (filed, not done here)

1. `packages/cli/src/commands/ci/init.ts:107` — adopter-facing emitter, identical defect.
2. `persona-architecture-enforcer.yml` `paths: src/**` — dead PR path filter; fix at the persona source YAML.

### DIFF
--- a/.changeset/persona-workflow-concurrency-1867.md
+++ b/.changeset/persona-workflow-concurrency-1867.md
@@ -1,0 +1,17 @@
+---
+'@harness-engineering/cli': patch
+---
+
+Persona CI-workflow generator: split the generated `concurrency` group by event so a push to a
+trunk branch no longer cancels the previous commit's verification.
+
+The generator emitted `group: ${{ github.workflow }}-${{ github.ref }}` with an unconditional
+`cancel-in-progress: true`. A persona declaring an `on_commit` trigger generates `push:` to a trunk
+branch, where `github.ref` is constant across every commit — so each push cancelled the still-running
+run for the preceding commit, concluding `cancelled` rather than `failure` and alarming nothing.
+
+Generated workflows now use the shape established in #1865 for the hand-written workflows:
+push keys the group on `github.sha` (per commit, never cancelled), while `pull_request` keeps the
+per-ref group with cancellation on, so PR supersession and PR runner spend are unchanged.
+
+Refs #1867.
--- a/.github/workflows/persona-architecture-enforcer.yml
+++ b/.github/workflows/persona-architecture-enforcer.yml
@@ -19,8 +19,8 @@ on:
   schedule:
     - cron: 0 6 * * 1
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 permissions:
   contents: read
 jobs:
--- a/.github/workflows/persona-codebase-health-analyst.yml
+++ b/.github/workflows/persona-codebase-health-analyst.yml
@@ -13,8 +13,8 @@ on:
     - cron: 0 6 * * 1
   pull_request: {}
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 permissions:
   contents: read
 jobs:
--- a/.github/workflows/persona-documentation-maintainer.yml
+++ b/.github/workflows/persona-documentation-maintainer.yml
@@ -17,8 +17,8 @@ on:
     branches:
       - main
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 permissions:
   contents: read
 jobs:
--- a/.github/workflows/persona-entropy-cleaner.yml
+++ b/.github/workflows/persona-entropy-cleaner.yml
@@ -12,8 +12,8 @@ on:
   schedule:
     - cron: 0 6 * * 1
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 permissions:
   contents: read
 jobs:
--- a/.github/workflows/persona-graph-maintainer.yml
+++ b/.github/workflows/persona-graph-maintainer.yml
@@ -15,8 +15,8 @@ on:
     branches:
       - main
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 permissions:
   contents: read
 jobs:
--- a/.github/workflows/persona-performance-guardian.yml
+++ b/.github/workflows/persona-performance-guardian.yml
@@ -16,8 +16,8 @@ on:
   schedule:
     - cron: 0 6 * * 1
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 permissions:
   contents: read
 jobs:
--- a/.github/workflows/persona-task-executor.yml
+++ b/.github/workflows/persona-task-executor.yml
@@ -18,8 +18,8 @@ on:
       - main
       - develop
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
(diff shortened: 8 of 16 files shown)


chadjw:
## cicd-fleet VERIFY note (independent — not the lane's self-report)

**Verdict: verified.** All 3 OSes green (ubuntu/macOS/windows, Node 22), 14 pass / 0 fail. The 2 `skipping` jobs (`comprehension-refresh`, `refresh-baselines`) are main-only write-back jobs correctly skipped on a PR.

### The concurrency expression is proven valid at runtime, not just asserted

An invalid `concurrency` expression fails at **workflow-parse time**, so the strongest available proof is that the regenerated workflows actually ran on this PR's head (`3b21aa05`). Four of them did, all green:

`Codebase Health Analyst` · `Documentation Maintainer` · `Performance Guardian` · `Task Executor`

That is real execution of the new `${{ github.event_name == 'pull_request' && github.ref || github.sha }}` group and the expression-valued `cancel-in-progress`, not merely a unit-test assertion about a string.

**What is still unproven here:** this PR exercises only the `pull_request` branch of the expression. The `push` branch — the per-commit group that is the actual fix — cannot be exercised until this lands on `main`. The first two rapid merges after landing are the confirming signal: both commits should keep their own persona runs instead of the earlier one flipping to `cancelled`.

### Independently confirmed

- Plan artifact at `docs/changes/persona-workflow-cancel-1867/plans/2026-09-08-persona-workflow-cancel-plan.md`, tracing a real `harness-workflow-audit` run.
- All 7 regenerated files carry the fix consistently; `pnpm generate:persona-workflows:check` passes, so generator and output agree.
- Regression tests guard **both** directions (push not cancelling, PR still cancelling) — and were shown non-vacuous by failing against the pre-fix generator.
- **Scope honoured:** `packages/cli/src/commands/ci/init.ts` is absent from the diff (verified as an empty diff, not merely claimed). The adopter-facing emitter was deliberately left alone and filed as #2050.
- `Refs #1867`, not `Closes` — the fleet does not auto-close; the human decides on landing.

### One correction to the dispatch brief, in the lane's favour

The brief stated all seven persona workflows trigger on `push`. **Only four do** — `architecture-enforcer`, `documentation-maintainer`, `graph-maintainer`, `task-executor`. The other three are PR/schedule-only. #1867's own title ("4 generated workflows") was accurate and the brief overstated it. The observed evidence is unaffected: the three cancelled runs cited (`34239670105`, `34239670102`, `34239670064`) all fall inside that set of four. Applying the fix to all seven anyway is correct — uniform generated output, and harmless for the three that never take the push path.

### Base freshness

CI was gathered against pinned base `db9c6739da57f6dd38a2e9d83bf94a1bfce9ca7b`; `main` has since advanced to `fef03ac5e`. This repository has **no `required_status_checks` rule**, so nothing enforces up-to-date-before-merge and this green is **provisional against a moving main** — re-check before landing.

*`cicd-fleet` never merges. This PR is handed back unmerged for human review.*

Issue #2046, opened 2026-09-08
Wall-clock elapsed-time assertions in seven unit tests are a flake class — one already reddened macOS CI by 5.5ms


chadjw:
## Summary

Seven unit tests assert a **wall-clock millisecond budget** with `expect(elapsed).toBeLessThan(N)`. On shared CI runners these are coin flips, not assertions. One of them has already taken `main` red.

## Proven red (not hypothetical)

`packages/core/tests/security/injection-patterns.test.ts:280` failed macOS CI on 2026-09-06, run `34030345441` (main push, SHA `56630d21`):

```
FAIL tests/security/injection-patterns.test.ts > scanForInjection > performance > scans 10KB input in under 100ms
##[error]AssertionError: expected 105.49008300000014 to be less than 100
Test Files  1 failed | 472 passed | 1 skipped (474)
```

It missed by **5.5 ms**. Nothing about the code under test changed; the runner was slightly busier.

## The full class at `db9c6739d`

| Site | Budget |
| --- | --- |
| `packages/core/tests/security/injection-patterns.test.ts:280` | 100 ms — **proven red** |
| `packages/core/tests/state/state-manager-cache.test.ts:39` | **50 ms** |
| `packages/core/tests/state/state-manager-cache.test.ts:110` | **50 ms** |
| `packages/core/tests/review/fan-out.test.ts:81` | 500 ms |
| `packages/core/tests/state/learnings.test.ts:883` | 8000 ms |
| `packages/cli/tests/skill/dispatcher.perf.test.ts:59` | 500 ms |
| `packages/cli/tests/skill/dispatcher.perf.test.ts:73` | 1000 ms |

The two 50 ms sites are the most exposed. Their own inline comment concedes the problem while keeping the defect:

```js
expect(elapsed).toBeLessThan(50); // CI runners can be slow; cache hit is still orders of magnitude faster than disk
```

That comment states the correct *intent* — "orders of magnitude faster than disk" — and then encodes it as an absolute wall-clock number, which is exactly the part that cannot hold on a shared runner.

## Why this is a distinct class

This is **not** #1827 (random bind ports, now fixed and guarded) and **not** #620 (parallel-unsafe tests under the pre-push gate, marked done). Those concern resource collisions and concurrency isolation. This class is a *timing* dependency: the assertion is a function of runner load, so it fails without any state being shared or any resource colliding. It needs its own remediation.

## Proposed fix

[...]

PR #2085, opened 2026-09-08
fix(core-tests): give each SC2 property seed its own timeout budget (windows CI red)

chadjw:
## Why

`main` @ `8e7a59800` is RED. CI run [`34271802379`](https://github.com/Intense-Visions/harness-engineering/actions/runs/34271802379), job `build-and-test (windows-latest, 22)`:

```
 FAIL  tests/state/event-sourcing/snapshot.property.test.ts > SC2 — reduce(events) === readSnapshot() (property) > holds over 200 randomized seeds on both the computed and fresh-hit paths
Error: Test timed out in 60000ms.
 ❯ tests/state/event-sourcing/snapshot.property.test.ts:121:3
```

The file reported **66925ms** for that one test. ubuntu and macOS passed the identical pinned seeds. The three commits after `8e7a59800` are all `[skip ci]`, so the current tip `738b296c0` has never run CI.

## Root cause

`snapshot.property.test.ts:121` was **one** `it()` looping `seed = 1..200`. **vitest budgets timeouts per test, not per unit of work**, so aggregating 200 seeds of real filesystem I/O under a single `it()` turned the suite's total wall-clock into an implicit, machine-speed-dependent assertion sitting under a fixed 60s ceiling.

The logic is fully deterministic — the mulberry32 seeds are pinned. Nothing about the *correctness* of the property varies by platform; only the *speed* does.

I/O volume under that one budget, measured by replaying the pinned PRNG offline:

| Quantity | Value |
| --- | --- |
| Seeds | 200 |
| Real `emitEvent` calls (each an `appendFileSync`) | **2974** |
| Mean events per seed | 14.9 |
| Max events in one seed | 30 |

Each `emitEvent` is several syscalls, not one. `emitEvent` (`log.ts:199`) does `getStateDir` (`existsSync`) → `mkdirSync` → `readTailSeq` → `appendFileSync`, and `readTailSeq` (`log.ts:46`) is `existsSync` plus a **full `readFileSync` of the growing log, re-read on every append** (its INV-2 contract). Per seed there is additionally a `mkdtempSync`, a `loadEvents`, two `readSnapshot`, a `materialize` (`writeFileSync` + `renameSync`), and a recursive `rmSync`. That puts the file on the order of **~18,000 filesystem syscalls inside a single 60s budget**.

Implied per-op cost: **~3.7ms on windows-latest** (66925ms / ~18k) against **~0.1ms on this macOS host**. That ratio is the entire failure.

`vitest.config.ts` already sets `testTimeout: 60_000` and `fileParallelism: false` on win32, so the Windows leg was *already* serialized and *already* on a raised ceiling — the budget had simply been consumed by aggregation. The Windows leg also runs `pnpm test -- --continue` (no coverage), so coverage instrumentation is not a factor.

**This is a defect in the test's structure, not in `snapshot.ts` or `log.ts`.** `readTailSeq`'s re-read-per-append is a deliberate, documented multi-writer correctness choice (INV-2), and 30 events is well inside its intended range. No product change is warranted, and none is made here.

## Fix

Split the single `it()` into `it.each(SEEDS)` — one case per seed.

- Each seed carries **its own** 60s budget, so no single test's runtime is load-bearing. vitest applies no per-file timeout, so the ~67s Windows file no longer fails.
- `afterEach` reclaims **that seed's** temp dir immediately, instead of holding all 200 live until the end of the run.
- Failures name the seed (`… — seed 137`) instead of one opaque aggregate failure.
- Cases stay **sequential**. `it.concurrent` would genuinely cut wall-clock but is **unsound** here: `buildLog` mutates `process.env.HARNESS_EVENT_WRITER_ID` and the module-level writer-id cache, so concurrent seeds would clobber each other's writer identity.

**All 200 seeds still run on every OS.** Nothing is skipped, retried, reduced, or platform-gated — the file's header ("A failing seed is a real reducer/snapshot bug to fix — not a test to weaken") is respected. The reporter now shows 200 named cases, which is what makes full coverage *visible* rather than merely asserted.

Alongside, the four `if (!x.ok) return;` narrowing guards become an `expectOk` helper that **throws**. At the base those `return`s were unreachable (the preceding `expect(...).toBe(true)` throws first), but a `return` inside a per-seed case would end that case **green having asserted nothing**. `expectOk` removes that silently-green path by construction and surfaces the underlying `Result.error` message instead of a bare `expected false to be true`.

### Rejected: just raise `testTimeout`

That leaves the same wall-clock assertion in place and merely moves the cliff further away. The budget would still be consumed by 200 aggregated seeds, so the next slower runner — or the next seed added — reintroduces the identical failure with no new diagnostic signal. That is the shape #2046 already tracks; adding to it is not a fix. Splitting removes the coupling instead of widening it.

## Measurements

| | Before (aggregate `it()`) | After (`it.each`, 200 cases) |
| --- | --- | --- |
| windows-latest CI | **66925ms → FAIL** (60000ms budget) | pending this PR's CI |
| macOS local, run 1 | 2.46s total / 1.93s in tests | 505ms total / 351ms in tests |
| macOS local, run 2 | 836ms / 562ms | 514ms / 359ms |
| macOS local, run 3 | 552ms / 388ms | 506ms / 358ms |
| Slowest single test | the whole file (388–1930ms) vs 60000ms budget | **5ms** vs its 60000ms budget (~12,000x margin) |
| Tests reported | 1 | **200 passed (200)** |

Verified after the fix:
- 3 consecutive runs green (above).
- Every seed 1..200 named **exactly once**; 0 missing from the verbose reporter.
- 0 cases above 1000ms.
- Full event-sourcing suite: 14 files, 350 tests, green.
- Full orchestrator suite (touched by the pre-push gate): 2926 tests green.
- `pnpm lint`, `pnpm format:check`, and core `tsc --noEmit` all clean.

**Mutation check — the property still bites.** Injecting a mutant into `materialize` (`meta.lastSeq + 1`) turns **all 200 cases RED**, each naming its own seed:

```
⎯⎯⎯⎯⎯⎯ Failed Tests 200 ⎯⎯⎯⎯⎯⎯
AssertionError: seed 1 fresh-hit path mismatch: expected { schemaVersion: 2, …(4) } to deeply equal { schemaVersion: 2, …(4) }
AssertionError: seed 2 fresh-hit path mismatch: ...
```

This proves the refactor did not hollow out the property, and demonstrates the per-seed naming win. Mutant reverted; `snapshot.ts` is byte-identical to base.

## Assumptions made

- **The 66925ms is dominated by filesystem latency, not a hang or a behavioural difference.** Supported by ubuntu/macOS passing the same deterministic seeds and by the per-op cost the ratio implies (~3.7ms, ordinary for Windows CI with real-time AV scanning of a temp directory). Not directly measured — see below.
- **vitest applies no per-file timeout**, only per-test (`testTimeout`) and per-hook (`hookTimeout`). If a per-file ceiling existed, the split would not help.
- The Windows leg runs without coverage instrumentation, so v8 coverage load is not a contributing factor.

## What I could NOT verify

- **I could not reproduce the Windows-only timeout locally.** This work was done on macOS — the leg that *passed* in CI. Locally the file runs in 0.4–1.9s against a 60000ms budget (~30–150x margin), so the failure cannot manifest here. The evidence that the budget is the binding constraint is the measured Windows/macOS ratio plus the syscall accounting above, **not** a local reproduction.
- **Post-fix behaviour on windows-latest is confirmed only by this PR's own CI run**, not locally.

## Notes for the reviewer

- **Test-only change.** No file under `packages/*/src/` is touched, so no changeset is required — `scripts/check-changesets.mjs` `SKIP_FILE` excludes `tests/` and `*.test.ts`, and the gate confirmed "No publishable package changes detected."
- **`pnpm format:check` passes over the whole tree.** One unrelated file did trip it during pushing: the pre-push security-ledger step rewrites the tracked `packages/cli/.harness/security/timeline.json` with raw `JSON.stringify(…, 2)` output rather than prettier's formatting, and `.prettierignore`'s `.harness/security/` entry is root-anchored so it does not cover the `packages/cli/` copy. **I did not touch or commit that file** — I reverted the gate's local write. It looks like a latent repo papercut (every pushing contributor re-dirties it), worth a separate issue; flagging rather than silently fixing unrelated drift.
- Plan artifact from the `harness-debugging` pipeline: `docs/changes/snapshot-property-per-seed-budget/plans/2026-09-08-snapshot-property-per-seed-budget-plan.md`.

Refs #2046 — same defect class (wall-clock as an implicit assertion), different shape: there the assertion is explicit in the test body, here it was implicit in the runner's per-test budget. Deliberately **not** `Closes`, since this is not one of the seven wall-clock assertions that issue enumerates.

Base: `738b296c0a50cb9890474124d466f38903e468e2`

### DIFF
--- a/.harness/comprehension/packages/core/tests/state/event-sourcing/_module.md
+++ b/.harness/comprehension/packages/core/tests/state/event-sourcing/_module.md
@@ -1,11 +1,10 @@
 ---
 schemaVersion: 1
 module: 'packages/core/tests/state/event-sourcing'
-sourceHash: 'edd1fb3dce26bc6a5abd979fde6fcef738ff195dee8cc23c40e9af852ea21466'
-compiledAt: '2026-08-28T01:22:11.105Z'
+sourceHash: '15f25bd8bf9b3960de4f5601683e9821cd7fbeefaa8ed96332912a4cbd828bf5'
 compiler: { static: '1.0.0', semantic: '1.0.0' }
-model: 'claude-haiku-4-5-20251001'
-semantic: present
+model: null
+semantic: absent
 members:
   [
     'concurrency.test.ts',
@@ -21,28 +20,6 @@ members:
     'writer-id.test.ts',
   ]
 ---
-
-## Summary
-
-The event-sourcing test suite validates a concurrent, event-sourced state architecture where all state changes are immutable events. It tests schema validation, multi-process safe writing with blob spilling for large payloads, lane state machine transitions with guarded edges, snapshot materialization, event ordering, and projections. Core invariants enforce no event loss, no sequence repetition, correct blob handling under concurrency, snapshot consistency, transition table enforcement, and dependency/evidence/force guards for lane state changes.
-
-## Invariants
-
-- No event loss under concurrent N-process writes; each (seq, writerId) pair is unique
-- Global event ordering: sorted by (seq asc, writerId asc) regardless of write sequence
-- Blob spill succeeds under concurrent writes; identical payloads collapse to single content-addressed blob
-- Missing blob doesn't abort load; affected event is dropped, valid events remain readable
-- Snapshots materialize correctly, reflecting all state transitions (Truth #7)
-- Lane transition table: allowed edges (planned→claimed→in_progress→in_review→done); rework via in_review↔in_progress; any→{blocked,canceled}
-- Off-table transitions rejected unless forced with actor + reason
-- Entering in_progress requires all dependent tasks in done lane
-- Entering done requires non-empty evidence array
-- EventSchema validates type-payload match; StoredEventSchema recognizes new types on disk
-- Projections are additive: audit events don't change coreState/lanes byte-representation
-- Writer ID unique per process via HARNESS_EVENT_WRITER_ID; monotonic seq counters per writer
-- Snapshot staleness tracking; materialize() refreshes on-demand within debounce window
-- Event log/blob dir structure invariant (paths normalized; Windows posix-normalized)
-- Scope envelope fields (stream, session) can be undefined; schema accepts both
 
 ## Interface Contract
 
--- a/docs/changes/snapshot-property-per-seed-budget/plans/2026-09-08-snapshot-property-per-seed-budget-plan.md
+++ b/docs/changes/snapshot-property-per-seed-budget/plans/2026-09-08-snapshot-property-per-seed-budget-plan.md
@@ -1,0 +1,91 @@
+# Plan: split the SC2 snapshot property test into per-seed cases
+
+**Date:** 2026-09-08 · **Driver:** windows-latest CI timeout on `main` @ `8e7a59800` (run `34271802379`, job `102214934241`) · **Tasks:** 4 · **Integration Tier:** small · **Refs:** #2046 (same defect class, different shape)
+
+Produced by the `harness-debugging` pipeline (investigate → analyze → hypothesize → fix). Debug session: `.harness/debug/resolved/snapshot-property-win-timeout.md`.
+
+## Symptom
+
+```
+ FAIL  tests/state/event-sourcing/snapshot.property.test.ts > SC2 — reduce(events) === readSnapshot() (property) > holds over 200 randomized seeds on both the computed and fresh-hit paths
+Error: Test timed out in 60000ms.
+ ❯ tests/state/event-sourcing/snapshot.property.test.ts:121:3
+```
+
+The file reported `66925ms` for that single test. ubuntu and macOS were green; only `build-and-test (windows-latest, 22)` failed.
+
+## Root cause
+
+`packages/core/tests/state/event-sourcing/snapshot.property.test.ts:121` is **one** `it()` that loops `seed = 1..200`, and vitest's timeout budget is **per test**, not per unit of work. Every seed's real filesystem I/O is therefore billed to a single 60s budget, which makes the total wall-clock of 200 seeds an implicit, machine-speed-dependent assertion. The logic is fully deterministic (pinned mulberry32 seeds) — nothing about the _correctness_ of the property varies by platform. Only the _speed_ does.
+
+The I/O volume under that one budget, measured by replaying the pinned PRNG (`scratchpad/count-events.mjs`, reproduced in the session log):
+
+| Quantity                                          | Value    |
+| ------------------------------------------------- | -------- |
+| Seeds                                             | 200      |
+| Real `emitEvent` calls (each an `appendFileSync`) | **2974** |
+| Mean events per seed                              | 14.9     |
+| Max events in one seed                            | 30       |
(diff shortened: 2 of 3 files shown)


chadjw:
## CI evidence — the Windows leg now passes the test that reddened `main`

**All three OS legs are green** (`build-and-test` ubuntu / macOS / windows-latest), plus `bench`, `harness`, `required-review`, `changeset-check`, and 4x `enforce`.

From the windows-latest job log:

```
 ✓ tests/state/event-sourcing/snapshot.property.test.ts (200 tests) 9930ms
```

200 named cases, all passing, against the `66925ms` **timeout failure** on `main` @ `8e7a59800`. Note this is the *file* total, and vitest budgets per test — no individual seed comes close to its own 60s ceiling.

### Honest caveat on that number

I did **not** predict a drop from 66925ms to 9930ms and I can't fully attribute it. Two candidates: this runner was simply less loaded, and/or reclaiming each seed's temp dir in `afterEach` (instead of letting 200 accumulate for the whole run) genuinely cut NTFS/AV pressure. **The fix does not depend on that speedup** — it depends on the budget now scaling with the unit of work rather than the batch. Even at the original 66925ms the file would pass, because no single test is load-bearing any more.

### First windows attempt failed on an unrelated pre-existing flake

Attempt 1 of the windows leg failed, but **not** on this test — `snapshot.property.test.ts` passed there too. It failed in a different package (`@harness-engineering/cli`) on a Windows `rmdir` race in an `afterAll`:

```
 FAIL  tests/comprehension/comprehend-smoke.e2e.test.ts > harness comprehend — E2E smoke (static, no LLM)
Error: EBUSY: resource busy or locked, rmdir 'C:\Users\RUNNER~1\AppData\Local\Temp\comprehend-smoke-N81amG'
 ❯ tests/comprehension/comprehend-smoke.e2e.test.ts:60:15
```

That run reported `Tests  8698 passed | 29 skipped` — **zero failing assertions**; the only failure was the teardown hook. The re-run passed. This looks like a genuine latent Windows flake worth its own issue (same family as this PR — Windows filesystem timing surfacing in tests — but a distinct defect: an unreleased handle at `afterAll` rather than a timeout budget). Flagging rather than folding it into this PR.

### `Reconcile audit exceptions` is red for reasons unrelated to this PR

```
[uncovered] Active advisory GHSA-4r6h-5v86-94p3 has no auditExceptions entry.
  - GHSA-4r6h-5v86-94p3 (high, liquidjs)
```

This branch touches **no** `package.json` and **no** lockfile (test + docs only). The advisory is active on the default branch — GitHub flagged it on push (dependabot alert 109) — and the sibling branch `fc3/cicd-merge-race-survivors` fails the identical check at the same timestamp, while branches from earlier today passed. So it became active on `main` in the interim and will red every PR until someone adds an `auditExceptions` entry or bumps `liquidjs`. Out of scope here; not silently fixed.

Issue #2093, opened 2026-09-08
pnpm.overrides security floors go silently inert under a non-pinned or pnpm 10+ toolchain


chadjw:
`package.json` carries a `pnpm.overrides` block of **27 dependency security floors** (`hono >=4.12.34`, `undici ^7.29.0`, `qs >=6.16.0`, `postcss >=8.5.23`, `protobufjs ^7.6.3`, `fast-xml-parser >=5.7.0`, `body-parser >=2.3.0`, `ip-address >=10.3.1`, …).

Every `pnpm` invocation in this repo now emits:

```
[WARN] The "pnpm" field in package.json is no longer read by pnpm. The following
keys were ignored: "pnpm.neverBuiltDependencies", "pnpm.overrides".
```

So the entire block is inert. This is deliberate-looking configuration that silently stopped being enforced — `pnpm-workspace.yaml:9-10` documents the move *into* `package.json` on the grounds that "pnpm 8.x reads that field but not the `overrides` key in this workspace file", and `packageManager` is pinned to `pnpm@8.15.4`.

## Measured blast radius — latent, not active

Checked all 27 floors against the resolved versions in `pnpm-lock.yaml`:

| | count |
|---|---|
| satisfied by current resolution | 20 |
| **violated right now** | **0** |
| not present in lockfile | 7 |

Nothing is currently below its floor, because the lockfile was generated while the overrides were still honoured. **The exposure is on the next resolution** — a `pnpm install` that re-resolves, a dependency bump, or a lockfile regeneration can now drop a package below its security floor with no error and no diff to review.

`neverBuiltDependencies: ["cpu-features"]` is ignored by the same warning.

## Why this belongs to the "reports success without running" class

Tracked as #2084. A security control that emits a warning on every invocation and enforces nothing is indistinguishable, in CI output and in review, from one that is working.

## Not the cause of the current advisories

To be precise: this does **not** explain the six active advisories on `Reconcile audit exceptions`. The three `hono` advisories affect versions *above* the `>=4.12.34` floor, so honouring the override would not have prevented them.

## Suggested fix

Move the overrides to the location the current pnpm reads (per the linked pnpm settings docs), or pin the toolchain to a pnpm that honours `pnpm.overrides`. Either way, add a check that fails when the block is present but ignored — the failure mode here is silence.

Found during a `fleet-command` conductor run; verified at `origin/main` @ `738b296c0`.

chadjw:
[...]

PR #2092, opened 2026-09-08
security(deps): bump vitest 4.1.5 -> 4.1.11 to clear GHSA-82fw-gwwq-j7x9

chadjw:
Clears **[GHSA-82fw-gwwq-j7x9](https://github.com/advisories/GHSA-82fw-gwwq-j7x9)** / `CVE-2026-84373` (**moderate**, published 2026-09-08T20:46:45Z): *"Vitest: Path Traversal / Arbitrary File Read via `@vitest/mocker` Redirect Mock"*. Affected `vitest` and `@vitest/mocker` `>= 2.1.0, < 4.1.11`; first patched `4.1.11`.

Companion to #2090 (liquidjs). Refs #2087. **Deliberately a separate PR** — #2090 is 2 files and already all-OS green; this one re-resolves the test runner and deserved its own signal.

## Lockfile only — no manifest change

Every manifest already declares `^4.1.5` and `4.1.11` is in range, so **nothing outside `pnpm-lock.yaml` changed**: all **28** `specifier:` fields in the lockfile still read `^4.1.5`. No `overrides` pin, no `auditExceptions` entry.

No changeset — `node scripts/check-changesets.mjs` reports *"No publishable package changes detected"*, because `vitest` is a devDependency everywhere. Adding one would have forced a spurious version bump on packages whose published contents are unchanged.

## Blast radius: the `@vitest/*` family, and nothing else

`@vitest/coverage-v8` and `@vitest/ui` pin an **exact** peer on vitest (`coverage-v8@4.1.11` peers `vitest: 4.1.11`), so the family has to move together. Every package entry that moved:

| Moved | Reason |
| --- | --- |
| `vitest` (3 peer-variants), `@vitest/coverage-v8`, `@vitest/ui`, `@vitest/expect`, `@vitest/mocker`, `@vitest/pretty-format`, `@vitest/runner`, `@vitest/snapshot`, `@vitest/spy`, `@vitest/utils` | 4.1.5 -> 4.1.11 |
| `tinyglobby@0.2.15` **removed** | vitest 4.1.11 wants `^0.2.15`; pnpm reused the `0.2.17` already present for tsup, leaving 0.2.15 unreferenced |

`typescript`, `algolia`, `nan`, `canary-test-cli` and `tsup` are **untouched** — verified by grepping the diff for them (0 hits).

[...]

Issue #2087, opened 2026-09-08
Newly-published liquidjs advisory GHSA-4r6h-5v86-94p3 reds the blocking audit-exceptions check on EVERY open PR; patched version is already in range


chadjw:
**Every open PR in this repo is currently red on the blocking `Reconcile audit exceptions` check, for a reason unrelated to its own diff.**

## What happened

GitHub published **[GHSA-4r6h-5v86-94p3](https://github.com/advisories/GHSA-4r6h-5v86-94p3)** / `CVE-2026-69222` at **2026-09-08T18:10:35Z** — roughly three hours before this was filed.

- **Severity:** high
- **Summary:** *LiquidJS: Uncontrolled Resource Consumption in `join` filter allows template authors to bypass `memoryLimit` and crash the process*
- **Affected:** `liquidjs <= 10.27.1`
- **First patched:** `liquidjs@10.27.2`

`.github/workflows/audit-exceptions.yml` (added by #1324) reconciles every active advisory against an `auditExceptions` register in `package.json`. That register is **absent entirely** — `package.json` has no `auditExceptions` key, so there are zero entries and coverage is zero. The job is deliberately blocking (no `continue-on-error`), so the moment any advisory goes active, it fails:

```
audit-exceptions: 1 active advisory, 0 register entries.
Active advisories:
  - GHSA-4r6h-5v86-94p3 (high, liquidjs)

audit-exceptions FAILED:
  [uncovered] Active advisory GHSA-4r6h-5v86-94p3 has no auditExceptions entry.
  Fix the dependency, or add an entry with a justification and an `expires` date.
```

The flip is visible in the workflow's own run history — nothing in the repo changed between these:

| Time (UTC) | Branch | Result |
| --- | --- | --- |
| 18:32 | `fc2/craft-naming-orch-core-2001` | success |
| 18:46 | `fc2/craft-naming-core-rehearsal-2009` | success |
| **18:10** | — | **advisory published** |
| 21:10 | `fc3/cicd-snapshot-property-timeout` | **failure** |
| 21:12 | `fc3/cicd-merge-race-survivors` | **failure** |

## The fix is almost certainly a one-line lockfile bump

[...]

PR #2102, opened 2026-09-09
security(deps): raise the hono override floor to >=4.13.5, clearing three advisories

chadjw:
Clears **three** advisories in one bump — all published 2026-09-08, all affecting `hono < 4.13.5`, first patched **4.13.5**:

| Advisory | Sev | Summary |
| --- | --- | --- |
| [`GHSA-crvj-82cr-hjcx`](https://github.com/advisories/GHSA-crvj-82cr-hjcx) | moderate | Query parser reads parameters after the URL fragment, causing cache-key and proxy interpretation differentials |
| [`GHSA-g6gw-c38x-mqfc`](https://github.com/advisories/GHSA-g6gw-c38x-mqfc) | moderate | Unbounded dot-notation nesting in `parseBody()` can cause memory exhaustion |
| [`GHSA-gqvv-2mrq-wpjv`](https://github.com/advisories/GHSA-gqvv-2mrq-wpjv) | moderate | Incomplete fix for CVE-2026-39408: `toSSG()` still writes files outside the output directory |

Fourth in the series with #2090 (liquidjs) and #2092 (vitest). Refs #2087.

## Why this touches `package.json` and not only the lockfile

This one is **not** the same shape as #2090/#2092, and the difference matters.

hono's resolution is **not governed by the manifest range**. The root `package.json` `pnpm.overrides` block carries **27 security floors**, one of which is `hono: '>=4.12.34'` — and the lockfile records *that override* as the importer's specifier:

```yaml
hono:
  specifier: '>=4.12.34'      # <- the override, not dashboard's ^4.12.18
  version: 4.13.1
```

I tried the lockfile-only technique from the previous two PRs first: raising `packages/dashboard/package.json` from `^4.12.18` to `^4.13.5` and re-resolving. **The lockfile did not move** — the override supersedes the manifest specifier entirely, and 4.13.1 already satisfies `>=4.12.34`.

So the floor is the only lever. This raises it to the first patched version. That is the mechanism this repo **already uses for exactly this purpose** — it is an update to an existing pin, not a new one.

### :warning: This contradicts #2093 — `pnpm.overrides` is NOT ignored

#2093 reports the whole `pnpm.overrides` block is silently ignored, with `[WARN] The "pnpm" field in package.json is no longer read by pnpm`. **That does not reproduce with this repo's pinned toolchain**, and the evidence here is direct:

- `packageManager` pins **pnpm@8.15.4** (via corepack). At that version, `pnpm install --lockfile-only` emits **no such warning** — I grepped for it specifically.
- The lockfile's own `overrides:` block is present with all 27 entries, and the importer specifier above proves the override is what is being applied.
- **Decisive:** changing *only* the floor `>=4.12.34` -> `>=4.13.5` moved hono `4.13.1` -> `4.13.7`. An ignored field cannot move a resolution.

The warning in #2093 is real but almost certainly came from a **newer pnpm** (pnpm 10 dropped reading the `pnpm` field in favour of `pnpm-workspace.yaml`). Under the pinned 8.15.4 the overrides are load-bearing. Worth correcting there before someone migrates the block on the strength of it — or, more dangerously, concludes the 27 security floors are already inert.

## What moved

pnpm resolved **4.13.7** — the newest satisfying the new floor, which is how every other floor in this block behaves (`>=4.12.34` had likewise landed on 4.13.1). 4.13.7 is a **zero-dependency** package with unchanged `engines.node: >=16.9.0` and is not deprecated.

Blast radius is exactly hono plus the `@hono/node-server@2.0.11` peer-variant key that names it:

```
-  /@hono/node-server@2.0.11(hono@4.13.1):   +  /@hono/node-server@2.0.11(hono@4.13.7):
-  /hono@4.13.1:                             +  /hono@4.13.7:
```

Nothing else moved. No changeset — `scripts/check-changesets.mjs` reports *"No publishable package changes detected"*.

## Evidence

- **Class:** `advisory-match`. **Strength:** `lockfile-only`.
- **Evidence cleared / no new finding**, gate's own script run against both trees at base `738b296c0`:

| Tree | Rows | Distinct GHSAs |
| --- | --- | --- |
| base `738b296c0` | 8 | 6 |
| this branch | 5 | **3** (all three hono advisories gone) |

Strictly better than base; nothing introduced.

**hono is a runtime dependency of `packages/dashboard`**, so 4.13.1 -> 4.13.7 can carry real behaviour change. All-OS `build-and-test` results at the head SHA are reported in a follow-up comment rather than assumed.

Base: `738b296c0a50cb9890474124d466f38903e468e2`. **Not merged, no auto-merge.**

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
       "cpu-features"
     ],
     "overrides": {
-      "hono": ">=4.12.34",
+      "hono": ">=4.13.5",
       "@grpc/grpc-js": "^1.14.4",
       "@hono/node-server": ">=2.0.10",
       "postcss": ">=8.5.23",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,7 +8,7 @@ neverBuiltDependencies:
   - cpu-features
 
 overrides:
-  hono: '>=4.12.34'
+  hono: '>=4.13.5'
   '@grpc/grpc-js': ^1.14.4
   '@hono/node-server': '>=2.0.10'
   postcss: '>=8.5.23'
@@ -326,7 +326,7 @@ importers:
         version: link:../types
       '@hono/node-server':
         specifier: '>=2.0.10'
-        version: 2.0.11(hono@4.13.1)
+        version: 2.0.11(hono@4.13.7)
       '@tailwindcss/typography':
         specifier: ^0.5.19
         version: 0.5.19(tailwindcss@4.2.2)
@@ -337,8 +337,8 @@ importers:
         specifier: ^13.0.6
         version: 13.0.6
       hono:
-        specifier: '>=4.12.34'
-        version: 4.13.1
+        specifier: '>=4.13.5'
+        version: 4.13.7
       lucide-react:
         specifier: ^1.8.0
         version: 1.8.0(react@18.3.1)
@@ -2475,13 +2475,13 @@ packages:
       yargs: 17.7.2
     dev: true
 
-  /@hono/node-server@2.0.11(hono@4.13.1):
+  /@hono/node-server@2.0.11(hono@4.13.7):
     resolution: {integrity: sha512-bjD221KPLoJTWUwso1J6fGKiTXEUFedG/s0visavY4zakFPkeGURMRNly+FhBHs7T8Dz4qHaZIMX9ZoJHSJtKA==}
     engines: {node: '>=20'}
     peerDependencies:
-      hono: '>=4.12.34'
+      hono: '>=4.13.5'
     dependencies:
-      hono: 4.13.1
+      hono: 4.13.7
     dev: false
 
   /@humanfs/core@0.19.2:
@@ -2742,7 +2742,7 @@ packages:
       '@cfworker/json-schema':
         optional: true
     dependencies:
-      '@hono/node-server': 2.0.11(hono@4.13.1)
+      '@hono/node-server': 2.0.11(hono@4.13.7)
       ajv: 8.18.0
       ajv-formats: 3.0.1(ajv@8.18.0)
       content-type: 1.0.5
@@ -2752,7 +2752,7 @@ packages:
       eventsource-parser: 3.0.6
       express: 5.2.1
       express-rate-limit: 8.3.2(express@5.2.1)
-      hono: 4.13.1
+      hono: 4.13.7
       jose: 6.2.2
       json-schema-typed: 8.0.2
       pkce-challenge: 5.0.1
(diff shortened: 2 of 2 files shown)


chadjw:
## All-OS result at `ede991cf040c84f542c1ee67ab1fd3f31121a8fe`

| Job | Result |
| --- | --- |
| `build-and-test (ubuntu-latest, 22)` | **SUCCESS** (attempt 1) |
| `build-and-test (windows-latest, 22)` | **SUCCESS** (attempt 1) |
| `build-and-test (macos-latest, 22)` | FAILURE attempt 1, **SUCCESS** attempt 2 |

hono is a **runtime** dependency of `packages/dashboard`, so the macOS red is documented rather than quietly re-run.

### The macOS failure cannot have been caused by this change

```
FAIL packages/burn tests/scan.test.ts > abstention > does not suppress an elevated status on a thin sample
AssertionError: expected 'EARLY' to be 'CRITICAL'
```

**Structural proof, not a statistical one:** `packages/burn` declares `"dependencies": {}` — zero runtime dependencies — and its devDependencies are only `@types/node`, `@vitest/coverage-v8`, `tsup`, `typescript`, `vitest`. Both `pnpm --filter @harness-engineering/burn why hono` and `... why js-yaml` return **nothing**. There is no path by which a hono bump can reach that test.

What it *is*: a wall-clock/week-boundary test. It pins `week_reset` to `{ weekday: utcIsoWeekday(now), time: '00:00', tz: 'UTC' }` and then asserts on spend attributed to the current week. This run executed at **2026-09-09T00:04Z** — four minutes past UTC midnight, right on the reset boundary it configures. A transcript line written "5 minutes ago" falls on the far side of that boundary, so the spend lands outside the week being evaluated and the status degrades `CRITICAL` -> `EARLY`, which is exactly the observed assertion.

It passed **12/12** locally on an untouched control worktree, and passed on ubuntu and windows in the same CI run.

**Not fixed here** — a correctness defect in another domain. Flagged for whoever owns `packages/burn`; the durable fix is to inject a fixed clock rather than deriving `week_reset` from `new Date()` at test time.

### The dashboard suite itself was green

Worth stating explicitly, since `packages/dashboard` is the actual hono consumer: its **144 tests passed (144)** on the failing macOS attempt. The only red was in a package that cannot see hono.

### Remaining check

Everything is green except `Reconcile audit exceptions`, expected and explained in the description. Its log at this SHA reports **5 rows**:

```
GHSA-4r6h-5v86-94p3 (high, liquidjs)          <- #2090, unmerged
GHSA-82fw-gwwq-j7x9 (moderate, vitest)        <- #2092, unmerged
GHSA-82fw-gwwq-j7x9 (moderate, @vitest/mocker)
GHSA-2883-xcg3-v3hh (high, js-yaml)           <- #2103, unmerged
GHSA-2883-xcg3-v3hh (high, js-yaml)
```

**All three hono advisories are absent** — `GHSA-crvj-82cr-hjcx`, `GHSA-g6gw-c38x-mqfc` and `GHSA-gqvv-2mrq-wpjv` are gone. That is the direct confirmation the floor bump works in CI, not just locally.

PR #2128, opened 2026-09-09
fix(core): wire canary review detectors by bare name, not plugin-qualified

ahhrealmonster:
## Summary

The four canary review detectors auto-wired at `after:REVIEW` / `after:FINAL_REVIEW` have never actually wired. `planCanaryReviewDetectors` matches the bare detector names against the caller's installed-skill catalog with a plain `set.has()`, but canary exposes its skills plugin-qualified as `canary:canary-cassandra` — the same form harness already dispatches in `harness-test-advisor` (`canary:canary-review-test`). Every lookup missed, so all four were skipped in a canary-present project.

## Why it stayed invisible

The skip is *contractually silent*: a forward-wired default must never hard-halt, so a missing detector is dropped and counted in the denominator rather than raised.

On top of that, both the module contract and `harness-autopilot/SKILL.md` asserted:

> As of canary **5.12.0** the plugin ships **none** of the four ... So today, in a canary-present project, **all four are skipped and REVIEW proceeds normally**

That was true when written. Canary has shipped all four since **7.2.0**. So an agent observing `0/4` consulted the docs, found that documented as the expected state, and proceeded. The documentation actively explained away the bug it was sitting on.

## Changes

- **`bareSkillName`** reduces a catalog entry to the skill name it designates. Detector names contain no colon, so taking the final segment is lossless for them and collapses every qualification depth (`canary:canary-katana`, `canary:skills:claude-code:canary-katana:canary-katana`) onto the bare name.
- **Availability matching is prefix-insensitive** for all three shapes. Set and array catalogs match bare or qualified; the predicate shape, which cannot be enumerated, probes the bare name and the canonical `canary:` form. Wired hooks still dispatch the **bare** name, so nothing downstream changes.
- **Dedup and the `enabled: false` opt-out normalize identically.** Previously a project declaring `canary:canary-cassandra` failed to match the default, so the detector was dispatched **twice** — and an `enabled: false` opt-out written in the qualified form was silently defeated. Same root cause, one layer down.
- **Stale version claims removed** from the module comment, the detector JSDoc, and the skill doc, with a note explaining why a canary version must not be restated there as current.

## Test-assertion change worth a look

`autopilot-skill-hooks.test.ts` asserted the skill doc **matches** `/5\.12\.0/` — it pinned the stale claim in place. It now asserts that string is **absent**, plus the durable contract (`prefix-insensitiv`, `denominator`). A doc test that pins a third-party version string is guaranteed to either go stale or block the correction.

## Tests

7 new cases: qualified, deep-qualified, mixed bare/qualified, and predicate-shaped catalogs; non-detector canary skills must not wire; and both dedup paths. They fail on `main` (7 failed / 23 passed) and pass here.

## Gates

| Gate | Result |
|---|---|
| build | 13/13 tasks |
| typecheck | clean |
| lint | 0 errors (1 pre-existing warning in an untouched file) |
| test — core | 5618 passed, 1 skipped |
| test — cli | 8730 passed, 3 skipped |

Changeset included (`@harness-engineering/core`, patch). Plugin artifacts regenerated for all five targets; `generate:plugin:check` is clean.

## Note for maintainers

I have not verified which availability form a live autopilot run actually supplies — only that both the bare-miss and the unknown-availability paths fail closed to the same silent `0/4`. If the real caller supplies nothing at all, this fix is necessary but not sufficient, and the caller needs to construct the catalog. Happy to follow up if you can point me at where that set is built.

Filed downstream as bop-clocktower/canary#837.

https://claude.ai/code/session_016Mg1EyLHfkgjoXGaZSTnqj


ahhrealmonster:
Pushed two commits: a follow-up fix plus the generated security-ledger entry.

**The prefix fix reached two of the three availability shapes, not all three.**

`toAvailabilityPredicate` normalizes an array or Set through `bareSkillName`, so qualification depth genuinely cannot matter for an enumerable catalog. A predicate cannot be enumerated, only probed — and it was probed with exactly two strings, the bare name and the two-segment `canary:` form.

A real Claude Code catalog reports neither. It uses the deep path `canary:skills:claude-code:canary-cassandra:canary-cassandra`. A predicate backed by that catalog therefore answered false for all four detectors, and all four were silently skipped: the original 0/4 defect surviving intact in the one shape the fix did not reach, invisible for exactly the same reason as before.

The deep form was already covered for the Set shape (test at line 317) and already named in `bareSkillName`'s own docstring, which is what made the predicate gap look like an oversight rather than a decision.

The probe list is now explicit as `CANARY_SKILL_PROBE_FORMS`, commented as being the actual coverage for that shape — a form omitted from it fails closed and says nothing, so the list is load-bearing.

Two new tests, both RED before the change:
- a predicate recognizing only the deep path for one detector, asserting the other three still report in `skipped` (so it cannot pass vacuously)
- a deep-path-only predicate wiring all four

Gates on the final tree: core typecheck clean, build 25/25, lint clean, core **5620 passed / 1 skipped**.

https://claude.ai/code/session_016hTwUC3QuD9wuKjPTBGr1z

ahhrealmonster:
## Verification: 0/4 → 4/4, evidenced by a run

The linked issue asks for the wiring to be demonstrated **by a run, not by reading
the code**. This is that evidence.

### Method

I built the availability set from the form a **live Claude Code session actually
reports**, not a hand-written ideal:

```
canary:skills:claude-code:canary-savant:canary-savant
canary:skills:claude-code:canary-blackhawk:canary-blackhawk
canary:skills:claude-code:canary-katana:canary-katana
canary:skills:claude-code:canary-cassandra:canary-cassandra
```

All four are present on disk in the installed plugin at canary **7.2.0**
(`agents/skills/claude-code/canary-{savant,blackhawk,katana,cassandra}`), which
independently refutes the "as of canary 5.12.0 the plugin ships NONE of them"
claim this PR removes.

I then called `planCanaryReviewDetectors(true, event, availability)` twice per
event — once with a `Set`, once with a predicate — against **two** module
versions: `origin/main` and this branch. Same catalog, same events, only the
module differs.

### Result

| module | `after:REVIEW` | `after:FINAL_REVIEW` |
|---|---|---|
| `origin/main` (pre-fix), Set | **wired 0/4**, skipped 4 | **wired 0/4**, skipped 4 |
| `origin/main` (pre-fix), predicate | **wired 0/4**, skipped 4 | **wired 0/4**, skipped 4 |
| this branch, Set | wired 4/4, skipped 0 | wired 4/4, skipped 0 |
| this branch, predicate | wired 4/4, skipped 0 | wired 4/4, skipped 0 |

The pre-fix run reproduces the reported defect exactly — all four detectors
silently skipped, in both availability shapes and both events — and this branch
closes it. The 0/4 row is the important half: it is the planted positive that
makes the 4/4 row mean something rather than being a green that was never able
to go red.

The probe files were scratch and are **not** committed; the committed suite
already covers the deep catalog form in four places. Existing tests re-run
green: 32 in `packages/core/tests/hooks/canary-review-hooks.test.ts`, 16 in
`packages/cli/tests/integration/autopilot-skill-hooks.test.ts`.

### What this does not prove

It exercises the planning function against a real catalog string, not a
full end-to-end autopilot invocation. It shows the name matching and the
dedup path are correct for every shape we know a catalog to take; it does not
prove some other caller constructs the availability set at all. That was
already called out as the open question in the issue, and the
unknown-availability default remains fail-closed.

### Note for the reviewer on the merge commit

This branch was `CONFLICTING`; the merge resolves one file,
`packages/cli/.harness/security/timeline.json`, which is append-only ledger
output. It was resolved by union rather than by taking a side:

- **`snapshots`** — unioned by `commitHash` and sorted by `capturedAt`, so all
  five scans survive in chronological order.
- **`findingLifecycles`** — both sides independently observed the *same*
  resolution of the three findings. The earliest mainline observation
  (`0663ca4`, 14:10) is kept, so the merge does not re-date when the two
  SEC-XSS-001 findings and the SEC-PTH-001 finding were actually fixed.

Worth knowing: before this merge the PR had an **empty check rollup** — CI had
never run on it at all. "No failing checks" was a zero denominator, not a pass.
It now runs, and is green.

chadjw (approved):

Issue #2141, opened 2026-09-13
Security timeline is written by `JSON.stringify` but format-gated by Prettier, so the pre-push hook blocks every push once a scan appends an entry


towlebooth:
## Symptom

`git push` is refused by the pre-push hook, on a file the hook itself just rewrote:

```
Checking formatting...
[warn] packages/cli/.harness/security/timeline.json
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
husky - pre-push script failed (code 123)
```

Nothing in the branch touched that file. The gate's own security scan appended a timeline entry earlier in the same hook run, and the format check at the end of that hook then rejected what the scan had written.

`prettier --write` on the file clears it — until the next scan appends another entry, at which point the push is blocked again. **The loop is not escapable by fixing the file**, because the thing that dirties it runs inside the hook that checks it.

## Root cause — two serializers, one file

`packages/core/src/security/security-timeline-manager.ts:82`:

```ts
writeFileSync(tmp, JSON.stringify(timeline, null, 2));
```

`JSON.stringify(value, null, 2)` and Prettier disagree on this file in two ways:

1. **Array expansion.** `JSON.stringify` puts every array element on its own line, unconditionally. Prettier collapses an array that fits inside `printWidth`. The committed file has

   ```json
   "findingIds": ["783a5c009a968228", "caebb5c38de36a6e", "8b93df1cff804a2e"]
   ```

   and the scanner rewrites it as

   ```json
   "findingIds": [
     "783a5c009a968228",
     "caebb5c38de36a6e",
     "8b93df1cff804a2e"
   ]
   ```

2. **No trailing newline.** `JSON.stringify` emits none; Prettier requires one. (The sibling writer at `packages/cli/src/templates/post-write.ts:44` does `JSON.stringify(config, null, 2) + '\n'` — so the codebase is already inconsistent about this.)

`.harness/security/timeline.json` is deliberately tracked — `post-write.ts:104` explicitly un-ignores it (`!security/timeline.json`) as the *"shared security trend ledger keyed by commit hash"* — so it is in scope for the pre-push format check over changed files.

## This is a solved problem elsewhere in the repo

`.prettierignore` already carries the exact rationale for three other generated artifacts:

> Machine-managed roadmap: **the deterministic serializer owns the format** of the generated aggregate and the per-row shards … Prettier would … **fight the regenerator forever** and trip the aggregate-drift check.

and

[...]

PR #2143, opened 2026-09-13
fix(scripts): generate-docs and the pre-push hook cannot run on Windows

towlebooth:
## Summary

`pnpm run generate-docs` fails on every Windows machine, and with it the pre-push hook that runs it. Three defects, all POSIX assumptions, each one hidden behind the last.

**They were only discoverable in sequence.** Fixing the `.bin` spawn let `generate-docs` run, which revealed the bare absolute `import()`, which let the CLI reference generate, which revealed that the generated docs are platform-specific, which turned the drift gate red — and only then did the hook reach `generate-tool-catalog.mjs`, which had the same `import()` bug again.

## 1. `execFileSync(node_modules/.bin/tsx, …)` → ENOENT

pnpm writes three shims there: `tsx` (a POSIX shell script), `tsx.CMD` and `tsx.ps1`. Windows cannot execute the extensionless one:

```
Error: spawn ...\node_modules\.bin\tsx ENOENT
```

Three scripts did this — `generate-agent-setup-prompt.mjs`, `generate-persona-workflows.mjs`, `generate-plugin.mjs`.

**The guard made it worse.** Each checked `existsSync(tsx)` first and exited with *"Missing tsx at … Run `pnpm install` first."* The POSIX shim **is** present, so the check passed, gave false confidence, and the real failure landed later and far less legibly than the message the guard existed to produce. Presence is not executability.

Fixed with `node --import tsx <entry>` rather than by resolving `tsx.CMD`: the `.CMD` route works but then needs `shell: true`, which needs every path quoted, on a repo whose checkout path may contain spaces. `--import` removes the shim from the picture — one code path, no shell, no quoting, identical on all three platforms. Output verified identical.

## 2. `await import(join(ROOT, …))` → "Received protocol 'c:'"

```
Only URLs with a scheme in: file, data, and node are supported by the default
ESM loader. On Windows, absolute paths must be valid file:// URLs.
```

The loader reads the drive letter as a URL scheme. POSIX absolute paths start with `/` and happen to work, which is why this survived. Three sites: two in `generate-docs.mjs` (hoisted to one constant, since both imported the same file) and one in `generate-tool-catalog.mjs`.

**And its error pointed at the wrong fix.** The catch reports *"CLI reference skipped (build CLI first: `pnpm build`)"* — so on Windows the docs silently lost their CLI reference and told the reader to run a build that had already succeeded. `docs/reference/cli-commands.md` now generates.

## 3. `path.join` in option defaults → platform-specific generated docs

`distortion`'s `DEFAULT_INPUT` / `DEFAULT_MODEL_OUT` were `path.join('.harness', 'metrics', …)`. They are option defaults, so they are printed in `--help` and embedded verbatim in the reference:

```diff
- `--input` — Observations JSONL (default: .harness/metrics/ablation-replays.jsonl)
+ `--input` — Observations JSONL (default: .harness\metrics\ablation-replays.jsonl)
```

A Windows regeneration therefore failed the reference-docs drift gate. **This could never surface before, because generate-docs could not run on Windows at all** — the platform bug was concealing the platform-specific output.

Fixed at the source as POSIX literals rather than by normalising separators in the generator, which would have to tell a path from any other backslash in the text. Node's fs accepts forward slashes on Windows, so nothing is lost. Changeset included (this one touches `packages/cli/src`).

## Verification

On Windows (Node 22.23.2, the repo's pinned major):

- `pnpm run generate-docs` — **exit 1 → exit 0**, and writes `cli-commands.md`, `mcp-tools.md`, `skills-catalog.md`, `agent-setup/prompt.md`
- the regenerated `docs/reference/` is **byte-identical** to the committed Linux-generated copies
- `pnpm run generate:tool-catalog:check` — **exit 1 → exit 0**
- `grep` confirms no bare absolute `await import(join(…))` or `.bin/tsx` spawn remains under `scripts/`
- the **full pre-push hook now passes**, which is how this branch was pushed

## Reported separately, not fixed here

While getting the hook to run I hit two more Windows issues that are out of scope for this PR:

- `pnpm run check:changesets` is `BASE_REF=origin/main node …` — POSIX env-var-prefix syntax, which cmd.exe rejects (`'BASE_REF' is not recognized…`), so the hook needs `npm_config_script_shell` pointed at a POSIX shell. Unfiled; happy to open one.
- #2141 — the security scanner writes `timeline.json` with `JSON.stringify(…, 2)` while Prettier format-gates it, so a scan that appends an entry blocks every push until the file is hand-formatted.

Both are the same family as these: tooling that assumes a POSIX shell or a Linux runner. None of them are caught by CI, because doc generation and the pre-push hook only ever run on Linux there.

### DIFF
--- a/.changeset/distortion-posix-path-defaults.md
+++ b/.changeset/distortion-posix-path-defaults.md
@@ -1,0 +1,19 @@
+---
+'@harness-engineering/cli': patch
+---
+
+Spell `distortion`'s path defaults as POSIX literals so the generated CLI reference
+is the same on every platform.
+
+`DEFAULT_INPUT`, `DEFAULT_MODEL_OUT` and `REFINEMENT_EVENTS` were built with
+`path.join('.harness', 'metrics', …)`. The first two are option defaults, so they are
+printed in `--help` and embedded verbatim in `docs/reference/cli-commands.md` — and a
+Windows regeneration emitted `.harness\metrics\…` where a Linux one emitted
+`.harness/metrics/…`, failing the reference-docs drift gate for whoever regenerated
+second.
+
+Node's fs accepts forward slashes on Windows, so the spelling is fixed at the source
+rather than teaching the doc generator to normalise separators, which would have to
+tell a path from any other backslash in the text.
+
+No behaviour change: the same files are read and written.
--- a/.harness/comprehension/packages/cli/src/commands/_module.md
+++ b/.harness/comprehension/packages/cli/src/commands/_module.md
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1
 module: 'packages/cli/src/commands'
-sourceHash: 'db809aaa5ba617b18554e552a550ca8ac8f1f34575b6b13f1b032b0ed012acd8'
+sourceHash: '7d5a4498cd843a59c95ff83c8967d60269ab9aefce86407cc92c17475b612750'
 compiler: { static: '1.0.0', semantic: '1.0.0' }
 model: null
 semantic: absent
--- a/.harness/comprehension/scripts/_module.md
+++ b/.harness/comprehension/scripts/_module.md
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1
 module: 'scripts'
-sourceHash: '781829badf43d391700cff6e41cfb322dc6c066299df5116fc88f5e10c3c3e33'
+sourceHash: 'f2819d9a6ff367b1663d4237ad14de7bb50ac9a29710744e716830ceb6fd08d2'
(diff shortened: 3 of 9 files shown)


chadjw:
Superseded by #2144. All 9 files here (distortion POSIX path defaults + the Windows tsx/pathToFileURL script fixes) are byte-identical to a subset of #2144's 13 files; #2144 additionally fixes the unreachable-registry false-positive in `harness update`. The only difference is the commands/_module.md sourceHash, which #2144 bumps further to cover both changes. Closing to avoid double-applying the same hunks and a _module.md hash conflict — everything in this PR ships via #2144.

Issue #317, opened 2026-05-14
[Bug] `harness update` prints "Update available" but never self-upgrades the CLI


ahhrealmonster:
## Summary

`harness update` reports that an upgrade is available but doesn't actually upgrade the CLI binary — it only refreshes hooks and regenerates platform integrations (slash commands, agent definitions). Running `harness update` repeatedly is a no-op for the version bump, which is misleading because the same command prints "All packages are up to date" immediately above "Update available: vX -> vY".

## Reproduction

Installed via `npm i -g @harness-engineering/cli` under nvm Node v25.2.1.

```
$ harness --version
2.3.0

$ harness update
i Checking for updates...
v All packages are up to date
v Refreshed 6 hooks (standard profile)
v Telemetry already configured

Regenerate slash commands, agent definitions, and knowledge graph? (Y/n) y
Generate for (G)lobal or (l)ocal project? (G/l) g
Generating slash commands...
  claude-code: 38 commands (0 new, 0 updated)
  gemini-cli: 38 commands (0 new, 0 updated)

Generating agent definitions...
  claude-code: 12 agents (0 new, 0 updated)
  gemini-cli: 12 agents (0 new, 0 updated)

Done.

Update available: v2.3.0 -> v2.3.1
Run "harness update" to upgrade.

$ harness --version
2.3.0
```

Running `harness update` again produces the exact same output — the "Run \`harness update\` to upgrade" hint loops forever. The actual upgrade only happens via `npm install -g @harness-engineering/cli@latest`.

## Expected

Either:

1. `harness update` actually performs the npm self-upgrade (preferred, matches what the banner promises), or
2. The CLI prints the real upgrade command (e.g. `npm install -g @harness-engineering/cli@latest`) instead of recommending `harness update`, and ideally drops the contradictory "All packages are up to date" line when a newer CLI version exists.

## Environment

- `@harness-engineering/cli` 2.3.0 -> 2.3.1 available
- Node v25.2.1 (nvm)
- macOS (Darwin 25.4.0)

## Impact

Users assume the CLI is current because `harness update` runs cleanly, but they're stuck on whatever version they first installed. This also bites in teams where the upgrade path isn't documented — happy to PR a docs update if useful while the behavior is being fixed.

chadjw:
I've got a fix for this landing soon

PR #2144, opened 2026-09-13
fix(cli): `harness update` called an unreachable registry "up to date"

### LINKED ISSUE #2139: `harness update` reports "All packages are up to date" when the registry query FAILS — a skipped check is indistinguishable from a current install
## Symptom

On the same machine, minutes apart, the CLI told me both of these:

```
$ harness <anything>
Update available: v12.1.0 -> v12.7.0
Run "harness update" to upgrade.

$ harness update
i Checking for updates...
v All packages are up to date
v Refreshed 9 hooks (standard profile)
v Telemetry already configured
```

The banner was right. The registry had `12.7.0`, the global install was `12.1.0`:

```
$ npm view @harness-engineering/cli version   → 12.7.0
$ npm ls -g @harness-engineering/cli          → @harness-engineering/cli@12.1.0
```

`npm install -g @harness-engineering/cli@latest` upgraded it correctly. **The update path was fine; the check in front of it was not.**

This cost real time: I believed `harness update`, concluded the CLI could not self-update, and went looking for the problem elsewhere. The user had also tried `harness update` earlier the same day and been told the same thing.

## Root cause — a rejected query is silently dropped, then emptiness is reported as success

`packages/cli/src/commands/update.ts:437` — `checkAllPackages()`:

```ts
const results = await Promise.allSettled(
  packages.map(async (pkg) => {
    const latest = await getLatestVersionAsync(pkg);
    const current = installedVersions[pkg] ?? null;
    return { pkg, current, latest, outdated: !current || current !== latest };
  })
);

const outdated: UpdateCheckResult['outdated'] = [];
for (const result of results) {
  if (result.status === 'rejected') {
    // Skip packages we can't query — don't block the whole update
    continue;
  }
  if (result.value.outdated) outdated.push(result.value);
}

return { hasUpdates: outdated.length > 0, outdated };
```

`getLatestVersionAsync` shells out to `npm view <pkg> dist-tags.latest` and throws on a timeout, a non-zero exit, or an empty response. When it rejects, that package contributes nothing to `outdated` — and at `:604` the caller reads the empty list as good news:

```ts
if (!hasUpdates) {
  logger.success('All packages are up to date');
```

**So "we could not check" and "you are current" produce identical output.** Any registry hiccup — timeout, proxy, offline, npm rate limit, corporate mirror — silently converts into a green success line. The failure mode is worse than a crash, because the user acts on it.

It also makes the banner's own advice unreliable: `Update available … Run "harness update" to upgrade` points at the one command that can quietly no-op.

Note this is a *different* defect from #317. That one was about `npm list -g` not seeing the running CLI, and its fix (the `CLI_VERSION` fallback at `:111`, plus `getInstalledPackages` always unshifting `CLI_PACKAGE`) is intact and working — the package list is never empty. The hole is downstream, in how a rejected *registry* lookup is accounted for.

## Suggested fix

Track rejections and refuse to claim success on an incomplete check:

```ts
const unreachable: string[] = [];
for (const result of results) {
  if (result.status === 'rejected') {
    unreachable.push(/* pkg name — needs carrying through the map */);
    continue;
  }
  ...
}
return { hasUpdates: outdated.length > 0, outdated, unreachable };
```

and at the call site, when `unreachable.length > 0`, say which packages could not be checked and exit non-zero rather than printing `All packages are up to date`. Distinguishing "checked, current" from "could not check" is the whole point — a checker that cannot reach the registry has not found anything.

(Carrying the package name through the rejection needs the `.map` to catch and return a tagged failure rather than throw, or a `Promise.allSettled` over `[pkg, promise]` pairs — either is fine.)

A regression test would mock `getLatestVersionAsync` to reject for one package and assert that `All packages are up to date` is **not** printed.

## What I could not verify

I am now on `12.7.0`, so I cannot reproduce the original state, and I did not capture which rejection fired at the time (no `--verbose`). So I cannot tell you *which* of the failure triggers I hit — only that the branch converts any of them into a false success, which is visible in the source above and sufficient on its own.

Environment: Windows 11, Node v24.11.0, npm global install, CLI 12.1.0 → 12.7.0.

towlebooth:
Closes #2139.

**Stacked on #2143** (the Windows toolchain fixes) — without those the pre-push hook cannot run on Windows at all, which is how this defect was found. Rebase or merge #2143 first; this branch contains those commits.

## The defect

A failed registry lookup contributed nothing to `outdated`, and the caller then read the empty list as good news.

`packages/cli/src/commands/update.ts` — `checkAllPackages()`:

```ts
for (const result of results) {
  if (result.status === 'rejected') {
    // Skip packages we can't query — don't block the whole update
    continue;
  }
  if (result.value.outdated) outdated.push(result.value);
}
return { hasUpdates: outdated.length > 0, outdated };
```

`getLatestVersionAsync` shells out to `npm view <pkg> dist-tags.latest` and throws on a timeout, a non-zero exit or an empty response. Offline, proxy, corporate mirror, registry throttle — any of those rejects, gets dropped, empties `outdated`, and prints:

```
v All packages are up to date
```

**So "we could not check" and "you are current" are the same output.** That is worse than a crash, because the user acts on it. It also makes the banner's own advice unreliable: `Update available … Run "harness update" to upgrade` points at a command that can silently no-op.

**Observed:** on the same machine, minutes apart, the banner said `v12.1.0 -> v12.7.0` while `harness update` said everything was current. `npm install -g @harness-engineering/cli@latest` upgraded it correctly — the update path was fine, the check in front of it was not. It cost me an hour and a wrong diagnosis, and the machine's owner had hit the same thing earlier the same day.

## The fix

Every failure is tagged with its package rather than being anonymous, and `UpdateCheckResult` gains `unreachable`. When any package could not be checked, `harness update` names each one and the reason, still prints the updates it *was* able to find, suggests the manual install command, and exits non-zero instead of claiming success.

**This is not #317.** That was `npm list -g` failing to see the running CLI, and its fix — the `CLI_VERSION` fallback at `:111` and `getInstalledPackages` always unshifting `CLI_PACKAGE` — is intact; the package list is never empty. The hole was downstream, in how a rejected *registry* lookup was accounted for.

## Tests

Two regression tests, **both confirmed to fail against the unfixed source**: reverting only `update.ts` reds exactly these two and leaves the other 49 green.

```
× does NOT claim everything is up to date when the registry cannot be reached
× names the package it could not check, and why
```

`errorSpy` is added to the suite's setup because `logger.error` writes to `console.error` while `logger.info`/`success` write to `console.log` — asserting on what the user was actually told needs both.

## Verification

On Windows, Node 22.23.2 (the repo's pinned major):

- `pnpm build` — 13/13 tasks
- the four update suites — **69 passed**
- `tsc --noEmit` — exit 0
- `eslint` — 0 errors
- `check:changesets` — OK, covers `@harness-engineering/cli`
- the full pre-push hook passes

Changeset included.

### DIFF
--- a/.changeset/distortion-posix-path-defaults.md
+++ b/.changeset/distortion-posix-path-defaults.md
@@ -1,0 +1,19 @@
+---
+'@harness-engineering/cli': patch
+---
+
+Spell `distortion`'s path defaults as POSIX literals so the generated CLI reference
+is the same on every platform.
+
+`DEFAULT_INPUT`, `DEFAULT_MODEL_OUT` and `REFINEMENT_EVENTS` were built with
+`path.join('.harness', 'metrics', …)`. The first two are option defaults, so they are
+printed in `--help` and embedded verbatim in `docs/reference/cli-commands.md` — and a
+Windows regeneration emitted `.harness\metrics\…` where a Linux one emitted
+`.harness/metrics/…`, failing the reference-docs drift gate for whoever regenerated
+second.
+
+Node's fs accepts forward slashes on Windows, so the spelling is fixed at the source
+rather than teaching the doc generator to normalise separators, which would have to
+tell a path from any other backslash in the text.
+
+No behaviour change: the same files are read and written.
--- a/.changeset/update-check-unreachable-registry.md
+++ b/.changeset/update-check-unreachable-registry.md
@@ -1,0 +1,20 @@
+---
+'@harness-engineering/cli': patch
+---
+
+`harness update` no longer reports "All packages are up to date" when it could not
+reach the registry.
+
+`checkAllPackages` collected results with `Promise.allSettled` and `continue`d past
+rejections, so a package whose `npm view` failed contributed nothing to `outdated` —
+and the caller then read an empty list as good news. Any npm hiccup (timeout, offline,
+proxy, throttle, non-zero exit, empty response) was silently rendered as a green
+success line, which is worse than a crash because the user acts on it. It also made
+the update banner's own advice unreliable: `Update available … Run "harness update"
+to upgrade` pointed at a command that could quietly no-op.
+
+Failed lookups are now tagged with their package and surfaced as `unreachable` on
+`UpdateCheckResult`. When any package could not be checked, `harness update` names
+each one and the reason, prints any updates it _was_ able to find, suggests the manual
+install command, and exits non-zero instead of claiming success. "We could not check"
+and "you are current" no longer produce the same output.
--- a/.harness/comprehension/packages/cli/src/commands/_module.md
+++ b/.harness/comprehension/packages/cli/src/commands/_module.md
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1
 module: 'packages/cli/src/commands'
-sourceHash: '7d5a4498cd843a59c95ff83c8967d60269ab9aefce86407cc92c17475b612750'
+sourceHash: 'ae84916a41f4c2c573c3fa903b245d28eef362565bd61af1386849d2a9bbd755'
 compiler: { static: '1.0.0', semantic: '1.0.0' }
 model: null
 semantic: absent
--- a/.harness/comprehension/packages/cli/src/commands/_module.md
+++ b/.harness/comprehension/packages/cli/src/commands/_module.md
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1
 module: 'packages/cli/src/commands'
-sourceHash: 'db809aaa5ba617b18554e552a550ca8ac8f1f34575b6b13f1b032b0ed012acd8'
+sourceHash: '7d5a4498cd843a59c95ff83c8967d60269ab9aefce86407cc92c17475b612750'
 compiler: { static: '1.0.0', semantic: '1.0.0' }
 model: null
 semantic: absent
--- a/.harness/comprehension/packages/cli/tests/commands/_module.md
+++ b/.harness/comprehension/packages/cli/tests/commands/_module.md
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1
 module: 'packages/cli/tests/commands'
-sourceHash: 'f616c8e87090edf43047bfe9275f85adf44caa85d2a58f03970f107de2b308a3'
+sourceHash: '5d9b41c96e468a9ab2689ac87a17172314ca309b316f2db04b8d2b5444dc2d7e'
 compiler: { static: '1.0.0', semantic: '1.0.0' }
 model: null
 semantic: absent
--- a/.harness/comprehension/scripts/_module.md
+++ b/.harness/comprehension/scripts/_module.md
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1
 module: 'scripts'
-sourceHash: '781829badf43d391700cff6e41cfb322dc6c066299df5116fc88f5e10c3c3e33'
+sourceHash: 'f2819d9a6ff367b1663d4237ad14de7bb50ac9a29710744e716830ceb6fd08d2'
 compiler: { static: '1.0.0', semantic: '1.0.0' }
 model: null
 semantic: absent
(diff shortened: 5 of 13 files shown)


ahhrealmonster (approved):
**Approving.** This is the right fix to the right bug, and the reasoning is documented where it will be read. CI is green on all three OSes. One note below and one mechanical blocker (a conflict in generated files) that I'm resolving rather than handing back.

### The core fix is correct

`checkAllPackages` swapped `Promise.allSettled` + `continue` for per-package `try`/`catch` with a `reached` discriminant, so a failed lookup now lands in `unreachable` instead of evaporating. That closes the actual defect: an empty `outdated` used to be indistinguishable from "every lookup failed," and the caller rendered both as `All packages are up to date`.

The exit-code change is the part that matters most. A scripted `harness update` previously read a network blip as "nothing to do" and exited 0; now an incomplete check exits non-zero and names each package with its reason. **A check that didn't run is not a check that passed** — this is exactly that distinction, and it's worth having mechanically.

Both new tests assert the right things: `not.toContain('All packages are up to date')` pins the defect itself rather than a proxy for it, and spying on `console.error` as well as `console.log` is necessary here (the two loggers write to different streams) and easy to get wrong.

### The three platform fixes are each correct and well-commented

- **`pathToFileURL` for dynamic `import()`** (`generate-docs.mjs`, `generate-tool-catalog.mjs`) — the ESM loader reads a Windows drive letter as a URL scheme and rejects it. The comment noting that the old `catch` reported "build CLI first: pnpm build" when the build was already done is the useful half: the docs silently lost their CLI reference and the message pointed at the wrong fix.
- **`node --import tsx` over `node_modules/.bin/tsx`** (3 generator scripts) — right call, and deleting the `existsSync` guard alongside it is right for the stated reason: the POSIX shim *is* present on Windows, so the guard passed and produced false confidence. Presence is not executability. `--import` avoids the `shell: true` + quoting path that resolving `tsx.CMD` would have required.
- **POSIX literals for `distortion`'s path defaults** — these are option defaults that get printed in `--help` and embedded verbatim in `docs/reference/cli-commands.md`, so `path.join` made the generated reference platform-specific and the drift gate failed for whoever regenerated second. Fixing the spelling at the source beats teaching the generator to tell a path from any other backslash. Node's fs accepts forward slashes on Windows, so there's no behaviour change.

### Note — scope

The PR title names only the `update` fix, but the change carries three independent fixes (registry reachability, the Windows generator-script breakage, and the `distortion` doc-drift spelling). The two changesets describe them honestly and each fix is small, well-isolated and well-argued, so I'm not going to make you split this — but a reader bisecting later will find the title under-describes the commit. Worth a fuller squash-merge subject.

### Conflict in generated files — resolving, not handing back

The branch conflicts with `main` on two files:

- `.harness/comprehension/packages/cli/src/commands/_module.md`
- `.harness/comprehension/packages/cli/tests/commands/_module.md`

Both conflicts are on the `sourceHash:` line only. That substrate is a derived cache, not an authoritative artifact, so this is a regeneration rather than a judgment call — no reason to bounce a green PR for it. I'm merging `main` in, regenerating the hashes, and confirming CI stays green before merge.

ahhrealmonster (approved):
**Re-approving.** Full CI is green on the merged head `23de33798` — 16 checks, 0 non-passing, including all three `build-and-test` OS jobs. Windows passing matters here specifically, since the Windows-only generator breakage is what two-thirds of this PR fixes.

My original approval was dismissed by my own push, so this restores it. Being explicit about what that push was, since it means this branch now contains a commit I authored and my approval is therefore not a clean outside review of 100% of the diff:

- Commit `23de33798` is a merge of `origin/main` that resolves the only two conflicts, both of which were on a single `sourceHash:` line in `.harness/comprehension/**/_module.md`.
- I regenerated those hashes with `harness comprehend --changed --static` rather than picking a side, because **neither side's hash was correct for the merged tree** — the merge combines main's source with this branch's edits, so both candidate hashes were stale and would have failed `harness comprehend --check`. The regenerated values (`7fdacb4b…`, `d06e55f7…`) differ from both.
- That same run also refreshed three units that were **already** source-stale on `main` before this branch existed (`packages/cli/src/mcp/tools`, `packages/core/src/comprehension`, `packages/core/tests/comprehension`). I reverted those out — an unrelated pre-existing gap doesn't belong in this PR — and filed them separately as #2181.
- I verified `main` has not touched any of this PR's source files since the merge base (`438563f68`), so the merge is purely additive and none of the reviewed logic changed under it.
- Post-resolution `harness comprehend --check` reports the same 3 stale units as clean `main` — this branch introduces zero new staleness.

So the substance I reviewed and approved is unchanged; my commit touches only two lines of a derived cache. Flagging it anyway rather than leaving it implicit.

The original review stands: the `unreachable` split is the right fix, the non-zero exit is the part that earns its keep, and all three platform fixes are correct and well-commented. Merging.

PR #2151, opened 2026-09-14
Name the stale-server cause instead of a deleted build chunk

chadjw:
## The failure

A long-lived MCP server whose CLI is upgraded underneath it keeps working — until a tool takes a code path it has not imported yet. Builds emit content-hashed chunks, so an upgrade **deletes** the old ones rather than replacing them in place. The lazy `import()` then resolves against a file that is gone:

```
Cannot find module '…/dist/dist-QI44KYEI.js'
  imported from '…/dist/chunk-GF46LTKE.js'
```

Both names are build artefacts that appear in no source file and no changelog. `dispatchTool` had no `try`/`catch`, so this reached the SDK raw.

## Why it needs translating rather than just surfacing

The message points every piece of evidence at the wrong culprit. The package on disk is **intact and at the latest published version**; every fresh `npx harness …` invocation works; only *some* tools fail, and they fail at first use rather than at startup. So it reads as a corrupt install — which is the one thing it is not. Reinstalling fixes nothing, because nothing is broken.

Observed in the field on 2026-09-14: this cost a full session's `manage_roadmap` access and **two wrong diagnoses** before the running process was identified as the stale party. The first was "the install is corrupt, reinstall it"; the second was "the install is stale, upgrade it." Both wrong — `npm ls -g` already showed the latest version. The tell was a `harness-mcp` process whose start time predated the `dist` mtime by three hours.

## The fix

`dispatchTool` recognises this one failure shape and answers with the cause and the remedy — **restart the server, do not reinstall** — plus the version actually installed on disk, read fresh rather than from the constant compiled into the stale build.

Everything else rethrows untouched. That restraint is the point: mislabelling an unrelated `ERR_MODULE_NOT_FOUND` would send someone to restart a server that was never the problem, trading one misleading error for another.

Recognition requires **both** that the missing specifier sits inside this build's own `dist/` **and** that it matches the hashed-chunk shape. Matching the shape rather than a list of known hashes is deliberate — the hashes change every build, so a list would be stale by construction. The conjunction rejects two things a looser check would swallow:

- a missing module in the user's own project
- a file we genuinely failed to ship — that is a packaging bug and wants a different report

This is a **recogniser, not a repair**. The process cannot reload its own deleted chunks; the only honest thing it can do is fail with an error that names the remedy.

## Verification

| | |
|---|---|
| New tests | 9, covering both prefixes, the `url`-absent fallback, and 6 must-not-match cases |
| Mutation proof | **4 mutations, 0 survivors** |
| `tsc --noEmit` | clean |
| `eslint` on changed files | clean |
| `src/mcp` + `tests/commands/mcp` | 312 passed |

Mutations run, each reverted and re-verified:

| Mutation | Result |
|---|---|
| drop the own-dist guard | RED 2/9 |
| drop the hashed-chunk guard | RED 2/9 |
| weaken the conjunction to a disjunction | RED 3/9 |
| drop the `ERR_MODULE_NOT_FOUND` check | RED 1/9 |

The disjunction mutation matters most: it passes every other test in the file, so the suite carries two cases that exist only to pin the `AND`.

## Pre-existing failure, not from this change

`src/mcp/tools/hermes-tools.test.ts` → *"returns Ok with empty matches when index is empty"* fails identically on pristine `origin/main` with this change fully reverted. It calls `handleSearchSessions` directly, never through `dispatchTool`, so it is structurally untouchable by this diff. Flagging rather than fixing — it wants its own change.

https://claude.ai/code/session_0171z7CJ3XXCDpeALQAsPRHj


ahhrealmonster (changes requested):
**Requesting changes.** The diagnosis is right and the module's rationale is excellent — this is a genuinely expensive failure to misread, and the comments earn their length. But the recogniser never fires on Windows, which is where the change is currently red, and there's a second path bug of the same family that CI structurally cannot catch.

---

### 1. Must-fix — the detector is dead on Windows (and this is why CI is red)

`packages/cli/src/mcp/stale-build.ts:47`

```ts
const HASHED_CHUNK = /\/(?:dist|chunk)-[A-Z0-9]{8,}\.js$/;
```

The leading `/` is hardcoded. `missingPath()` returns a **filesystem path** (via `fileURLToPath`), so on Windows that value is `D:\…\dist\dist-QI44KYEI.js` — backslash-separated. `HASHED_CHUNK.test(missing)` never matches, so `describeStaleBuildError` returns `null` for the exact input it was written to recognise.

Confirmed in CI: 4 of 9 tests in `stale-build.test.ts` fail on `windows-latest` with `expected null not to be null` ([run 34899828633](https://github.com/Intense-Visions/harness-engineering/actions/runs/34899828633/job/104766162595)), while ubuntu and macOS pass.

This isn't only a test-portability problem. A global npm package gets upgraded under a running server on Windows exactly as it does on POSIX, and tsup emits the same content-hashed chunks — so the feature is absent on one of the three platforms the repo tests, and the operator there still gets the opaque error this PR exists to replace.

Suggested shape — separator-agnostic, and it keeps every negative case in the test passing (those lean on the `ownDistDir` half of the conjunction, which is unaffected):

```ts
import path from 'node:path';

/** Basename only — the path separator differs by platform, the filename does not. */
const HASHED_CHUNK = /^(?:dist|chunk)-[A-Z0-9]{8,}\.js$/;
// ...
if (!HASHED_CHUNK.test(path.basename(missing))) return null;
```

While you're there, the `startsWith(ownDistDir())` comparison is worth normalising through `path.resolve` on both sides too — it happens to work today because both values come from `fileURLToPath`, but that's an invariant nothing pins.

### 2. Must-fix — missing changeset

`changeset-check` is red: `packages/cli` is publishable, so the repo requires a `.changeset/*.md` naming it and a bump level (`pnpm changeset`, or `--empty` if this genuinely shouldn't release — it should, it's a user-facing error-message change). The check cites incident #332 as the reason it's mechanical.

### 3. Should-fix — `installedVersion()` can never succeed in the shipped build

`packages/cli/src/mcp/stale-build.ts:105`

```ts
const pkg = require('../../package.json') as { version?: string };
```

That's correct relative to the **source** location (`packages/cli/src/mcp/` → `packages/cli/package.json`), which is why the tests are green. But tsup emits a **flat** bundle into `packages/cli/dist/`, so at runtime `import.meta.url` sits in `dist/` and the specifier resolves to `packages/cli/../../package.json` — one level too high. It throws, the `catch` returns `null`, and every real invocation prints:

> The version now installed on disk could not be read.

There's precedent for the correct depth in this same package — `packages/cli/src/version.ts:7`:

```ts
resolved = (require_('../package.json') as { version?: string }).version ?? '0.0.0';
```

So `'../package.json'` is the fix. Two things make this worth catching rather than shrugging at:

- The doc comment explicitly argues the version must be read **fresh from disk** rather than compiled in, and that reasoning is sound — but the silent fallback means the message loses the one piece of information the comment says is the point, in every install.
- The test suite runs from `src/`, so no test on any platform can observe this. Given the module's entire subject is a src-vs-dist layout skew, a dist-path bug slipping through is worth a regression guard — asserting that the resolved specifier is the package root (or just unit-testing a small `packageJsonPath()` helper) would pin it.

---

### Things I checked and am happy with

- The AND-not-OR conjunction test is exactly the right test to write, and the "flip one at a time" case is the kind of thing that actually survives a refactor.
- Catching at `dispatchTool`'s handler call is the right seam — it is the first place a lazy `import()` failure is catchable with context — and rethrowing anything unrecognised is the correct default.
- `[A-Z0-9]{8,}` matches esbuild's uppercase base32 hash alphabet.
- The `.url`-first / message-fallback ordering, and the comment explaining why, are right.

Happy to re-review as soon as the Windows path handling and the changeset are in — the substance here is good and I'd like to see it land.

